PR #14, opened 2022-02-16
fix: add validation when update pipeline and fix bug

bochengyang:
Because

- The validation of the model should also be checked during updating the pipeline
- Should have a definition of the status for mode

This commit

- fix: add validation when update pipeline and fix bug


Phelan164 (approved):
LGTM

PR #16, opened 2022-02-16
ci: make new opened issue to trigger workflow

xiaofei-du:
Because

- current workflow only triggers when a PR changes from draft -> ready for review status. We want to trigger the workflow whenever there is a new non-draft PR created.

This commit

- add trigger event `opened` and exclude draft PRs


xiaofei-du:
It worked. Please check the [added PR](https://github.com/orgs/instill-ai/projects/5/views/1).

PR #18, opened 2022-02-19
fix: correct codes during integration test

bochengyang:
Because

- found issues during integration test with vdp and need to be fixed

This commit

- correct codes during integration test


xiaofei-du (approved):
LGTM 👍

PR #19, opened 2022-02-19
fix: use direct instead of http for DS/DD in recipe

bochengyang:
Because

- we align the new term in the type of recipe if the trigger is from HTTP/gRPC

This commit

- use direct instead of http for DS/DD in recipe


xiaofei-du (approved):
LGTM 👍

xiaofei-du (approved):

PR #23, opened 2022-02-21
fix: change struct definition from private to public

bochengyang:
Because

- we need a public struct for reusing purpose

This commit

- change struct definition from private to public


pinglin (approved):
LGTM

PR #30, opened 2022-02-25
chore: use instill-ai/x module for zapadapter

pinglin:
Because

- we now have module `instill-ai/x` to share common package

This commit

- adopt `instill-ai/x/zapadapter` pacakge
- update `README.md`
- update `.gitignore`
- update codecov job

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -9,7 +9,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
 
-      - uses: actions/setup-go@v1
+      - uses: actions/setup-go@v3
         with:
           go-version: 1.17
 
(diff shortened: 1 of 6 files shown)


bochengyang:
@pinglin Does it make sense if we maintain our zapadapter for temporal in [instill-ai/x](https://github.com/instill-ai/x) because if we use sample-go as our reference, there is a possibility that file will be moved to another place since that repo is a sample repo

pinglin:
> @pinglin Does it make sense if we maintain our `zapadapter` for `Temporal` in [instill-ai/x](https://github.com/instill-ai/x) because if we use sample-go as our reference, there is a possibility that file will be moved to another place since that repo is a sample repo

Yeah, that temporalio repository name looks a bit risky. That makes sense. I will move this to `instill/x` and update this PR.

PR #31, opened 2022-03-04
test: fix go lint and add unit test

bochengyang:
Because

- pre-commit hook `golangci-lint` won't pass in each commit
- need add skeleton of unit test

This commit

- fix pre-commit golangci-lint
- add skeleton of unit test


pinglin (approved):
LGTM

PR #37, opened 2022-03-22
fix: fix response in snake case

pinglin:
Because

- response output standard follows snake_case to disambiguate REST and gRPC naming convention

This commit

- adopt protobuf@v0.1.5-alpha
- make codecov.yml unblock mode
- configure grpc-gateway to use snake_case


Phelan164 (approved):
LGTM

PR #40, opened 2022-03-24
refactor: keep codebase naming consistent

pinglin:
Because

- naming rule in the codebase is inconsistent and sometimes confusing 

This commit

- move `rpc` package from root to `pkg` folder and rename it to `handler`
- rename `model` in `pkg` folder to `datamodel` to avoid naming confusion with the pipeline model component
- remove plural form of struct name for following Golang naming convention


Phelan164 (approved):
LGTM

PR #59, opened 2022-07-08
chore: update protobuf message

Phelan164:
Because

- protobuf message change bytes to content
- error when trigger a pipeline with grpc stream method

This commit

- update required fields
- fix read file length when trigger a pipeline


xiaofei-du (commented):
LGTM 👍

PR #80, opened 2022-10-23
fix: pipeline trigger model hangout

### LINKED ISSUE #77: Server panic when triggering model-backend fails in sync mode
```
{"level":"error","ts":1666169847.2963574,"caller":"service/service.go:391","msg":"[model-backend] Error TriggerModel at 0th model instance models/yolov4/instances/v1.0-gpu: rpc error: code = FailedPrecondition desc = [handler] inference model error","stacktrace":"github.com/instill-ai/pipeline-backend/pkg/service.(*service).TriggerPipeline.func1\n\t/src/pkg/service/service.go:391"}
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x30 pc=0xf3750d]

goroutine 88 [running]:
github.com/instill-ai/pipeline-backend/pkg/service.(*service).TriggerPipeline.func1()
	/src/pkg/service/service.go:394 +0x32d
created by github.com/instill-ai/pipeline-backend/pkg/service.(*service).TriggerPipeline
	/src/pkg/service/service.go:381 +0x88c
```

### LINKED ISSUE #79: Trigger pipeline hang out
When request trigger models to model-backend timeout, the pipeline-backend do not finish goroutine and hangout (i.e could not continue to process other trigger requests)

Phelan164:
Because

- Trigger model timeout 5 seconds is not enough for making model inference
- Goroutines use sync.WaitGroup which does not decrease the counter due to exception errors

This commit

- Increase timeout to 30 seconds
- defer wg.Done() and end function when have any error
- close #79 
- close #77

### DIFF
--- a/pkg/service/service.go
+++ b/pkg/service/service.go
@@ -378,7 +378,10 @@ func (s *service) TriggerPipeline(req *pipelinePB.TriggerPipelineRequest, dbPipe
 	wg.Add(1)
 
 	var modelInstOutputs []*pipelinePB.ModelInstanceOutput
+	errors := make(chan error)
 	go func() {
+		defer wg.Done()
+
 		for idx, modelInstance := range dbPipeline.Recipe.ModelInstances {
 
 			ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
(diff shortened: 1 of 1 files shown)


pkg/service/service.go
```diff
@@ -405,29 +410,54 @@ func (s *service) TriggerPipeline(req *pipelinePB.TriggerPipelineRequest, dbPipe
...
 	case dbPipeline.Mode == datamodel.PipelineMode(pipelinePB.Pipeline_MODE_ASYNC):
 		go func() {
-			wg.Wait()
+			go func() {
+				wg.Wait()
+				close(errors)
+			}()
+			for err := range errors {
+				if err != nil {
+					logger.Error(fmt.Sprintf("[connector-backend] Error trigger model instance got error %v", err.Error()))
```
pinglin, line 456:
```suggestion
					logger.Error(fmt.Sprintf("[model-backend] Error trigger model instance got error %v", err.Error()))
```

pinglin (changes requested):

pinglin (approved):

PR #82, opened 2022-10-25
chore: support instance segmentation task

Phelan164:
Because

- support instance segmentation task

This commit

- add instance segmentation task output


pinglin (approved):
LGTM

PR #97, opened 2023-01-12
chore: update integration test

Phelan164:
Because

- create a model asynchronous

This commit

- update model preparation in integration test


pinglin (approved):
LGTM

PR #102, opened 2023-02-10
fix: replace fatal logs with error logs

xiaofei-du:
Because

- backend service should not be terminated because of connection to other services

This commit

- replace fatal logs with error logs


pinglin (approved):
LGTM

PR #103, opened 2023-02-13
chore: update test model for running on Triton server 22.12

Phelan164:
Because

- VDP update triton server to 22.12 which does not allow to import Tensor from triton_python_backend_utils (refer to the [issue](https://github.com/triton-inference-server/server/issues/4870)) 

This commit

- change import Tensor from c_python_backend_utils (temporary solution, maybe no need after Triton update about it)


pinglin (approved):
LGTM

PR #111, opened 2023-03-20
feat: adopt private/public apis for pipeline

Phelan164:
Because

- support private APIs used by other services internally

This commit

- add private APIs
- add gRPC test cases

Resolves INS-126


pinglin (approved):
LGTM

PR #130, opened 2023-03-31
chore: recognise owner based on `jwt-sub` in header if applicable

xiaofei-du:
Because

- if `jwt-sub` is in the request header, we use the UUID as the UUID of the request owner; if not, we use `owner-id` as the ID of the request owner.

This commit

- recognise the request owner based on the header `jwt-sub` or `owner-id`
- update tests


pinglin (approved):
LGTM

PR #134, opened 2023-04-05
feat(controller): add pipeline state monitoring with controller

heiruwu:
Because

- support pipeline state monitoring with controller

This commit

- add public /watch endpoint for querying pipeline state from etcd through controller
- use /watch in integration-test to reflect model current state


pinglin (approved):
LGTM

PR #138, opened 2023-04-14
chore: change owner from using id to uuid

donch1989:
Because

- The user ID can be changed now (in Instill Cloud), so using the owner ID is not reliable

This commit

- refactor owner of a pipeline to use the UUID, not the ID
- update protogen-go dependency


pinglin (changes requested):
Need to add integration test for the change

pinglin (approved):

PR #145, opened 2023-04-27
chore: check access_token in redis for trigger endpoint

donch1989:
Because

- Pipeline will get `access_token` from headers. We should use this to find out the owner of this pipeline.

This commit

- Query `access_token` from redis to get owner permalink

### DIFF
--- a/internal/resource/resource.go
+++ b/internal/resource/resource.go
@@ -2,9 +2,11 @@ package resource
 
 import (
 	"context"
+	"fmt"
 	"strings"
 	"time"
 
+	"github.com/go-redis/redis/v9"
 	"github.com/gofrs/uuid"
 	"github.com/gogo/status"
 	"github.com/instill-ai/pipeline-backend/pkg/constant"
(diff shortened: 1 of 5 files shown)


internal/resource/resource.go
```diff
@@ -63,11 +65,31 @@ func GetPermalinkUID(permalink string) (string, error) {
 }
 
 // GetOwner returns the resource owner
-func GetOwner(ctx context.Context, client mgmtPB.MgmtPrivateServiceClient) (*mgmtPB.User, error) {
+func GetOwner(ctx context.Context, client mgmtPB.MgmtPrivateServiceClient, redisClient *redis.Client) (*mgmtPB.User, error) {
 
+	// Verify if "authorization" is in the header
+	authorization := GetRequestSingleHeader(ctx, constant.HeaderAuthorization)
 	// Verify if "jwt-sub" is in the header
 	headerOwnerUId := GetRequestSingleHeader(ctx, constant.HeaderOwnerUIDKey)
```
xiaofei-du, line 73: do we still need to check `jwt-sub`, since it does not access JWT token now?

xiaofei-du (approved):

PR #148, opened 2023-05-05
fix(controller): fix rename id and execution order

heiruwu:
Because

- failing integration-test with pipeline renaming

This commit

- fix delete/update resource execution order
- fix id/newid


pinglin (approved):
LGTM

PR #182, opened 2023-06-28
chore(logging): update metric implementation

heiruwu:
Because

- support metric API in `mgmt-backend`

This commit

- write metric record directly to `influxDB` instead of instrumenting with `OpenTelemetry`
- update log format to accommodate future log querying


donch1989 (approved):
LGTM

PR #200, opened 2023-07-16
chore: rename trigger endpoint and implement GetOperation for TriggerAsync

donch1989:
Because

- simplify the naming of `trigger` endpoint
- implement a function for user to get the result of async pipeline

This commit

- rename `trigger` endpoint
- implement `GetOperation` endpoint for `TriggerAsync`


pinglin (dismissed):
LGTM

PR #210, opened 2023-08-01
chore(usage): update usage collection data format

heiruwu:
Because

- provide more detailed information in usage collection

This commit

- add usage collection trigger data and update usage data format

### DIFF
--- a/pkg/service/influx.go
+++ b/pkg/service/influx.go
@@ -1,7 +1,22 @@
 package service
 
-import "github.com/influxdata/influxdb-client-go/v2/api/write"
+import (
+	"context"
+	"fmt"
+	"time"
 
-func (s *service) WriteNewDataPoint(p *write.Point) {
-	s.influxDBWriteClient.WritePoint(p)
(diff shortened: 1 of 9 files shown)


pkg/service/influx.go
```diff
@@ -1,7 +1,22 @@
...
-func (s *service) WriteNewDataPoint(p *write.Point) {
-	s.influxDBWriteClient.WritePoint(p)
+	"github.com/instill-ai/pipeline-backend/pkg/utils"
+)
+
+// TODO: better naming for this file
+
+func (s *service) WriteNewDataPoint(ctx context.Context, data utils.UsageMetricData) {
+
+	s.redisClient.RPush(ctx, fmt.Sprintf("user:%s:trigger.trigger_time", data.OwnerUID), data.TriggerTime.Format(time.RFC3339Nano))
```
donch1989, line 15: The key is conflict with model-backend datapoint. I think the data will mess up together.

heiruwu: oh yes good catch, will update that

donch1989 (approved):
LGTM

PR #214, opened 2023-08-17
chore(usage): adopt new usage-client to include owner_uid in session

heiruwu:
Because

- avoid complex query for usage metrics

This commit

- adopt new usage-client to add `owner_uid` in session


donch1989 (approved):
LGTM

PR #235, opened 2023-09-06
chore(metric): fix pipeline release nil uuid value

heiruwu:
Because

- fix pipeline release nil uuid value

This commit

- fix pipeline release nil uuid value


donch1989 (approved):
LGTM

PR #262, opened 2023-09-26
chore: add dependencies for text extraction

praharshjain:
Because

- we need some dependencies for text-extraction

This commit

- adds the dependencies to dockerfile

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,6 +23,8 @@ RUN apt clean
 
 FROM gcr.io/distroless/base:nonroot
 
+USER nonroot:nonroot
+
 ARG SERVICE_NAME
 
 WORKDIR /${SERVICE_NAME}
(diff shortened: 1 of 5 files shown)


Dockerfile
```diff
@@ -9,6 +9,9 @@ COPY go.mod go.sum ./
 RUN go mod download
 COPY . .
 
+# install dependencies for text extraction (refer https://github.com/sajari/docconv)
+RUN apt-get install poppler-utils wv unrtf tidy
```
donch1989, line 13: We are using multi-stage docker build. I guess adding here is not enough?

donch1989:
@praharshjain please check the failed GA

Dockerfile
praharshjain: updated to use another base image

donch1989:
@praharshjain I think we also need those dependencies in Dockerfile.dev

donch1989:
I'm closing this PR since it is not working

PR #292, opened 2023-11-08
refactor(service): optimize recipe validation error message

donch1989:
Because

- the error message formats of recipe validation are not consistent

This commit

- optimize recipe validation error message
- add pipeline recipe validation before triggering


donch1989:
close https://github.com/instill-ai/community/issues/469

PR #311, opened 2023-12-05
feat: Improve error messages in connector execution

jvallesm:
Proof of concept of connector error UX improvement.

Because

- Connector errors provide an unfriendly UX.

This commit

- Extracts, if present, an end-user message from the pipeline trigger execution error.
- At the handler level, the error is inspected in search for an end-user message. It is used to set the gRPC status message. If not present, `err.Error()` is used.

CleanShot 2023-12-15 at 14 45 30

CleanShot 2023-12-15 at 14 44 32


pkg/middleware/interceptor.go
```diff
@@ -65,6 +65,20 @@ func AsGRPCError(err error) error {
 		return nil
 	}
 
+	// If it's already a status, respect the code.
+	// If, additionally, an end-user message has been set for the error, it has
+	// has priority over the status message.
+	if st, ok := status.FromError(err); ok {
```
jvallesm, line 71: Caught by integration tests 🥅 👋 ⚽

donch1989:
LGTM, can be merged after updating the connector package version

PR #368, opened 2024-01-23
chore(usage): add pipeline id in usage data

heiruwu:
Because

- pipeline_id is more informative in the context of grafana dashboard

This commit

- add pipeline_id in PipelineUsageData

resolves INS-3467


jvallesm (approved):
👍

PR #392, opened 2024-02-22
feat: use `mgmtPB.Owner` to embed the owner information in response

donch1989:
Because

- Originally, we used `owner_name` and `owner` (`structpb`) to embed owner information for Console. However, `owner` is a `structpb`, and it is not well-formatted. Since we already have dedicated messages for `User` and `Organization`, we should utilize these messages.

This commit

- Utilizes `mgmtPB.Owner` to embed the owner information in pipeline and connector response when `view != VIEW_BASIC`.
- Updates Connector and Operator packages.


jvallesm (approved):
👏 

The number of lines due to the `/integration-test/proto` directory scared me about reviewing this 😂 Is there a way we can dynamically fetch these e.g. on a `buildtest` command and add the directory to `.gitignore`?

donch1989:
> 👏
> 
> The number of lines due to the `/integration-test/proto` directory scared me about reviewing this 😂 Is there a way we can dynamically fetch these e.g. on a `buildtest` command and add the directory to `.gitignore`?

I think we should do that. It is very inconvenient to maintain the proto manually.

PR #409, opened 2024-03-06
feat: Implement component definition list filtering

jvallesm:
- **fix: avoid filtering out the tombstone components twice**
- **chore: return component definition description in endpoint**
- **feat: fetch component definitions from database**
- **feat: load component definitions on startup**

Because

This commit


jvallesm:
Wrong branch name

PR #410, opened 2024-03-06
feat: Introduce component definition list filtering

jvallesm:
Because

- Component definition list lacked filtering features that are needed for the component page (title, release stage, type).

This commit

- Adds a `component_definition_index` table in the database, which will hold the component definition fields we'll filter by.
  - This table should be used only to query the component definition UIDs and types. These can be used to fetch the definition from the inmem map, which is directly read from the `definitions.json` files and that is the source of truth for this entity.
- Leverages the filter SQL transpiler to add filtering capabilities to the component definition list endpoint.
- On service intialisation, if there are new components / components with a new version, their information will be converted to a `component_defintion_index` record, which will be upserted in the database.
  - This means that, from now on, when we modify the definition of a component we need to do a (minor) version bump.
  - ⚠️ This also means we shouldn't change the component UID from now on
  - ☝️ I'll document these 2 in the component / operator repos


donch1989 (approved):
LGTM

PR #412, opened 2024-03-08
feat: sort component list by score

jvallesm:
Because

- We want a way to feature / promote some components in the component page.

This commit

- Adds a map with the component ID -> component score, which will be used in the component definition table in DB to sort the filtered / paginated results.
- Because this is decoupled from the `definitions.json` file (reasoning [here](https://linear.app/instill-ai/issue/INS-3738/sort-component-lists-by-score#comment-9f1a1f51)), changes in these scores will trigger a database component definition update. The init logic that keeps the table up to date is refactored a little bit.

### DIFF
--- a/pkg/datamodel/datamodel.go
+++ b/pkg/datamodel/datamodel.go
@@ -401,6 +401,31 @@ type pbDefinition interface {
 	GetVersion() string
 }
 
+// FeatureScores holds the feature score of each component definition. If a
+// component definition is not present in the map, the score will be 0.
+//
+// This is implemented as an inmem map because:
+//   - We plan on having a more sophisticated rank in the future (e.g. likes,
+//     pipeline triggers where the pipeline uses this component).
+//   - We don't want to expose a detail in the component definition list endpoint
+//     (mainly used for marketing) in the component definition.
+//   - Having a unified place with all the scores provides a quick way to figure
+//     out the resulting order.
+//
+// TODO when a component major version changes, a new component with a
+// different ID will be introduced. We'll probably want to keep the score, so
+// it would be good that the new ID is xxx.v1 and that we use everything before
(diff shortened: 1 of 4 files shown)


jvallesm:
# QA ✅ 

With the following score map:

```go
var FeatureScores = map[string]int{
	"instill-model": 50,
	"openai":        40,
	"pinecone":      30,
	"numbers":       30,
	"json":          20,
	"redis":         20,
}
```

We have the following order:
```sh
curl --request GET \
  --url "localhost:8080/vdp/v1beta/component-definitions?page_size=25" \
  --header 'accept: application/json' | jq -c '.component_definitions[] | if .type == "COMPONENT_TYPE_OPERATOR" then .operator_definition  else .connector_definition end | .id'


  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 20675    0 20675    0     0   520k      0 --:--:-- --:--:-- --:--:--  531k
"instill-model"
"openai"
"numbers"
"pinecone"
"json"
"redis"
"airbyte-destination"
"bigquery"
"gcs"
"google-search"
"restapi"
"website"
"base64"
"start"
"end"
"image"
"text"
"hugging-face"
"stability-ai"
"archetype-ai"
```

(previous order was: `["instill-model","hugging-face","openai","stability-ai",...]`)

pkg/datamodel/datamodel.go
```diff
@@ -401,6 +401,31 @@ type pbDefinition interface {
 	GetVersion() string
 }
 
+// FeatureScores holds the feature score of each component definition. If a
```
donch1989, line 404: This makes sense. The score should be beyond the scope of the component definition. I think this is a good approach until we have the actual ranking data.

donch1989 (approved):

PR #426, opened 2024-03-19
fix: duplicated values in component list

jvallesm:
Because

- Some results were repeated across pages in the component definitions endpoint

CleanShot 2024-03-19 at 11 38 04

This happens because the results were sorted by `feature_score` at the database level before paginating. The rest of the results (score 0) were unsorted, which causes an arbitrary order in each query.

This commit

- Sorts the results by UID before paginating
- Additionally, it sorts them by release stage (given the same score, we'll want to have the more stable components first).

### DIFF
--- a/integration-test/pipeline/rest-component-definition.js
+++ b/integration-test/pipeline/rest-component-definition.js
@@ -45,18 +45,18 @@ export function CheckList() {
     });
 
     // Access non-first page.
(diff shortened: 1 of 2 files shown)


jvallesm:
# QA ✅ 
CleanShot 2024-03-19 at 11 46 51

donch1989 (approved):

PR #451, opened 2024-04-09
feat: add repository tests against DB

jvallesm:
Because

- Repository layer has no coverage. We tend to rely on integration tests to verify this package communicates with the database correctly but these are expensive and should verify how each layer interconnects rather than covering all the behaviours of each layer.
- I set up these tests for [INS-4081](https://linear.app/instill-ai/issue/INS-4081/credit-data-model) as it has no high-level entrypoint (yet). The credit data model is being moved to `mgmt-backend` but having this functionality set up in `pipeline-backend` is still useful.

This commit

- Adds a simple test in `pkg/repository` that runs against a test DB. For that:
  - A new database `pipeline_test` is created with the current migration in the local PostgreSQL database.
  - Config initialisation takes a config path (so tests can also read the main configuration).
  - The `coverage` github action is modified to include the database coverage, too.

# Notes

`make unit-test` is modified to 
- be renamed as `make coverage`
- take a flag `DBTEST` to allow / disable DB tests
- take a flag `HTML` display the coverage as HTML

- Previous `make unit-test` -> `make coverage HTML=true`
- Run tests locally with the DB package -> `make coverage DBTEST=true`
- ... and display coverage (and clean up coverage report) -> `make coverage HTML=true DBTEST=true`

### DIFF
--- a/.env
+++ b/.env
@@ -13,3 +13,7 @@ SOCAT_PORT=2375
 # container build
 DOCKER_BUILDKIT=1
 COMPOSE_DOCKER_CLI_BUILD=1
+
+# test
+TEST_DBHOST=localhost
+TEST_DBNAME=pipeline_test
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -9,6 +9,18 @@ jobs:
     permissions:
       contents: read
       pull-requests: write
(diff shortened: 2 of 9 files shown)


donch1989 (approved):
🥇

PR #478, opened 2024-05-09
feat: add tag system for pipeline

chuang8511:
Because

- we will use tag to filter

This commit

- create tag table & CRUD for tag
- add tag filter for pipeline in functions, but there isn't a caller yet.

### DIFF
--- a/pkg/db/migration/000014_init.up.sql
+++ b/pkg/db/migration/000014_init.up.sql
@@ -1,0 +1,18 @@
+BEGIN;
+
+CREATE TABLE IF NOT EXISTS public.tag(
+    uid UUID NOT NULL,
+    id VARCHAR(255) NOT NULL,
+    pipeline_uid UUID NOT NULL,
+    tag_name VARCHAR(255) NOT NULL
+    create_time TIMESTAMPTZ DEFAULT CURRENT_TIMESTAMP NOT NULL,
+    update_time TIMESTAMPTZ DEFAULT CURRENT_TIMESTAMP NOT NULL,
+    delete_time TIMESTAMPTZ DEFAULT CURRENT_TIMESTAMP NULL,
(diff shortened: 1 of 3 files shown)


donch1989:
@chuang8511
Please update the PR title.

pkg/db/migration/000014_init.up.sql
```diff
@@ -0,0 +1,18 @@
+BEGIN;
+
+CREATE TABLE IF NOT EXISTS public.tag(
+    uid UUID NOT NULL,
+    id VARCHAR(255) NOT NULL,
```
donch1989, line 5: Since the tag table won't be used standalone, I think we don't need another `uid` and `id` field. Keeping only `pipeline_uid`, `tag_name`, and timestamps should be enough.

pkg/repository/repository.go
```diff
@@ -67,6 +67,11 @@ type Repository interface {
 	GetNamespaceSecretByID(ctx context.Context, ownerPermalink string, id string) (*datamodel.Secret, error)
 	UpdateNamespaceSecretByID(ctx context.Context, ownerPermalink string, id string, secret *datamodel.Secret) error
 	DeleteNamespaceSecretByID(ctx context.Context, ownerPermalink string, id string) error
+
+	CreateNamespaceTag(ctx context.Context, tag *datamodel.Tag) error
```
donch1989, line 71: - `CreatePipelineTag(ctx, pipelineUID, tagName)`
- `DeletePipelineTag(ctx, pipelineUID, tagName)`
- `ListPipelineTags(ctx, pipelineUID)`

donch1989 (changes requested):

chuang8511:
@donch1989 Modified the comment parts.

PR #484, opened 2024-05-13
feat: return total pipeline count to frontend

chuang8511:
Because

- We want to let the users know the count of public pipeline

This commit

- get the pipeline count


chuang8511:
I will redo it

PR #488, opened 2024-05-16
chore: export component definition update logic

jvallesm:
Because

- Component definition update logic is duplicated in Cloud. This caused a bug when both versions became out of sync.

This commit

- Exports logic so changes are propagated to Cloud.


donch1989:
Hi @jvallesm , will this affect today's release?

PR #492, opened 2024-05-20
feat: support python code in components

chuang8511:
Because

- we want to use python lib in pdf component

This commit

- add python virtual machine in container

### DIFF
--- a/Dockerfile.dev
+++ b/Dockerfile.dev
@@ -10,10 +10,12 @@ WORKDIR /${SERVICE_NAME}
 ARG TARGETOS TARGETARCH K6_VERSION
 
 # Install Python, create virtual environment, and install pdfplumber
-RUN apt-get update && \
-    apt-get install -y python3 python3-venv && \
+RUN apt update && \
+    apt install -y python3 python3-venv && \
     python3 -m venv /opt/venv && \
-    /opt/venv/bin/pip install pdfplumber
+    /opt/venv/bin/pip install pdfplumber && \
+    apt install -y poppler-utils wv unrtf tidy tesseract-ocr libtesseract-dev && \
+    rm -rf /var/lib/apt/lists/*
 
 # air
 RUN --mount=target=. --mount=type=cache,target=/root/.cache/go-build --mount=type=cache,target=/go/pkg GOOS=$TARGETOS GOARCH=$TARGETARCH go install github.com/cosmtrek/air@v1.49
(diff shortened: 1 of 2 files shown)


chuang8511:
@donch1989 
I am curious about if we still use Dockerfile in pipeline-backend.
I searched in instill-core and pipeline-back, but I did not find where we execute it.

jvallesm:
> @donch1989 I am curious about if we still use Dockerfile in pipeline-backend. I searched in instill-core and pipeline-back, but I did not find where we execute it.

Without looking into much detail, I think the step they're used in is the image build GitHub Action.

chuang8511:
@jvallesm 
Thanks for your response! 
I think I understand and found the code! Thank you!!

chuang8511:
@jvallesm @donch1989 
Usually, when we modify Dockerfile rather Dockerfile.dev, how do you test to verify it?
I build the image by Dockerfile, but it seems the container will not be out of instill-core.
I am not quite sure about this part. Do you have some special skills to test it?

jvallesm:
> @jvallesm @donch1989 Usually, when we modify Dockerfile rather Dockerfile.dev, how do you test to verify it? I build the image by Dockerfile, but it seems the container will not be out of instill-core. I am not quite sure about this part. Do you have some special skills to test it?

I haven't faced this use case yet but I'd follow the regular dev env setup replacing `Dockerfile.dev` with `Dockerfile` in the `make build` step.

jvallesm (dismissed):
The changes look good.

chuang8511:
> I haven't faced this use case yet but I'd follow the regular dev env setup replacing Dockerfile.dev with Dockerfile in the make build step.

Yes, I did that before. 
But, it seems that I cannot isolate the container from instill-core & pipeline-backend.
It seems there is a small difference between Dockerfile & Dockerfile.dev, which makes me unable to do it.

jvallesm:
> > I haven't faced this use case yet but I'd follow the regular dev env setup replacing Dockerfile.dev with Dockerfile in the make build step.
> 
> Yes, I did that before. But, it seems that I cannot isolate the container from instill-core & pipeline-backend. It seems there is a small difference between Dockerfile & Dockerfile.dev, which makes me unable to do it.

I see, please address this question on Slack #dev-vdp or Discord as others might have experienced this before.

chuang8511:
Tested with Dockerfile, and it is ok now!

Dockerfile.dev
```diff
@@ -9,6 +9,14 @@ WORKDIR /${SERVICE_NAME}
 
 ARG TARGETOS TARGETARCH K6_VERSION
 
+# Install Python, create virtual environment, and install pdfplumber
+RUN apt update && \
+    apt install -y python3 python3-venv && \
+    python3 -m venv /opt/venv && \
+    /opt/venv/bin/pip install pdfplumber && \
+    apt install -y poppler-utils wv unrtf tidy tesseract-ocr libtesseract-dev && \
```
donch1989, line 17: We can merge this to L14

PR #497, opened 2024-05-30
feat: users can create and update tags by updating pipeline

chuang8511:
Because

- we want users can create tags for pipelines for better ux

This commit

- can create and update tags by update pipeline


pkg/service/pipeline.go
```diff
@@ -322,6 +322,34 @@ func (s *service) UpdateNamespacePipelineByID(ctx context.Context, ns resource.N
 		return nil, err
 	}
 
+	toUpdTags := toUpdPipeline.GetTags()
+
+	// We will need to revert it after the frontend is ready.
```
tillknuesting, line 327: nit; use TODO and be more specific why changes are needed and when

pkg/service/pipeline.go
```diff
@@ -322,6 +322,34 @@ func (s *service) UpdateNamespacePipelineByID(ctx context.Context, ns resource.N
...
+	if len(toUpdTags) > 0 {
+		currentTags := existingPipeline.TagNames()
+		toBeCreatedTagNames := []string{}
```
tillknuesting, line 330: would it make sense here to pre-allocate the slice with len(toUpdTags) for performance ?

chuang8511: @tillknuesting 
Thanks for telling me it! I did not know it.
I researched a little. 
It seems it will be better if expected length is close to actual length for the memory & performance.

However, there are 2 slices here, which are `toBeCreatedTagNames` & `toBeDeletedTagNames`.
`len(toBeDeletedTagNames)` + `len(toBeCreatedTagNames)` will be equal to `len(toUpdTags)`
It makes it a trade-off here.

But, in this part, maybe there won't be too many tags in this part.
So, I guess two ways will be similar. 
**And, I will modify it by pre-allocation.**

Thank you for telling me this again! 
I will also consider it for future development

chuang8511:
I am not sure why it happened...
But, it seems not so related to this PR.
And, I have tested before rebasing. I think it is ok to review.
So, I changed the status of this PR.
image

pkg/datamodel/datamodel_test.go
tillknuesting, line 1: if you have time let us add also a negative case

pkg/service/pipeline.go
```diff
@@ -322,6 +322,34 @@ func (s *service) UpdateNamespacePipelineByID(ctx context.Context, ns resource.N
 		return nil, err
 	}
 
+	toUpdTags := toUpdPipeline.GetTags()
+
+	// TODO: we will need to revert it after the frontend is ready.
```
tillknuesting, line 327: Please be specific in comments like this. Somebody else reading the comment does not necessarily know what "ready" means but would understand "when this feature x is implemented in the fronted". If you are planning to remove this by your self you can use "TODO chuang8511: remove this when the fronted supports updating tags of pipelines. 


 If its a task you are going to solve use "TODO

chuang8511: I take it out because it will be merged after the console is released.
- [ref](https://github.com/instill-ai/pipeline-backend/pull/497#issuecomment-2142506640)

chuang8511:
I am considering to add `go generate ./pkg/mock/generator.go` in CI process.
I will add it if I meet the similar situation.

tillknuesting:
> I am considering to add `go generate ./pkg/mock/generator.go` in CI process. I will add it if I meet the similar situation.
> 
> Situation is that the CI process failed because the latest code is not included in the mock types.

I think its fine to generate code at CI level to assert that it matches the checked in code but the test should be run against code that is checked into version control to have reproduce builds. Whats wrong with just generating the code and committing the files?

chuang8511:
> Whats wrong with just generating the code and committing the files?

I did not know the test file is because of test file itself or the function code.
I think it is also ok to generate the code and commit the files.
If we have re-generate the mock files in CI process, it helps us target the problems quicker.

donch1989:
> > Whats wrong with just generating the code and committing the files?
> 
> I did not know the test file is because of test file itself or the function code. I think it is also ok to generate the code and commit the files. If we have re-generate the mock files in CI process, it helps us target the problems quicker.

I think the main point is that we cannot guarantee that the code in the codebase is aligned with the generated code in the CI workflow. So if you want to generate the code in CI, you also need to check in the generated code into the codebase via CI.

chuang8511:
@donch1989 

> I think the main point is that we cannot guarantee that the code in the codebase is aligned with the generated code in the CI workflow. So if you want to generate the code in CI, you also need to check in the generated code into the codebase via CI.

Yes. I will do it in another branch if this is always annoyed us.
Should I rebase now?
Or, do we still wait for Console?

PR #508, opened 2024-06-07
chore: improve dependency injections

jvallesm:
Because

- One common problem in backend repos is the way dependencies are injected. Though communicating through interfaces, usually these are defined in the implementation package, making the client effectively import the implem package anyways.

This commit

- Removes `repository` dependencies from `service`. Further work can be done on this until we invert these dependencies having `repository` importing domain entities and errors from `service`. However, this would affect other packages (e.g. `worker`, which also imports `repository`) that are currently being modified by other features (e.g. streaming), so let's merge these well-bounded changes and achive saner injections step by step.
- The motivation for these changes was having a timeseries repository that can be used for token and credit metrics and, eventually, for logging. The groundwork for this will be laid in the cloud repos, where the dependencies are more manageable.


cmd/init/definitionupdater/updater.go
```diff
@@ -8,12 +8,13 @@ import (
 	"github.com/gofrs/uuid"
 	"github.com/launchdarkly/go-semver"
 
+	componentstore "github.com/instill-ai/component/store"
```
donch1989, line 11: Hi @jvallesm 

Let's group the package import in this way. 
1. Golang built-in packages
2. packages from others
3. packages from others with alias
4. packages from Instill AI
5. packages from Instill AI with alias

```
"github.com/gofrs/uuid"
"github.com/launchdarkly/go-semver"

"github.com/instill-ai/pipeline-backend/pkg/datamodel"
"github.com/instill-ai/pipeline-backend/pkg/logger"
"github.com/instill-ai/pipeline-backend/pkg/repository"

componentstore "github.com/instill-ai/component/store"
errdomain "github.com/instill-ai/pipeline-backend/pkg/errors"
pb "github.com/instill-ai/protogen-go/vdp/pipeline/v1beta"
```

It's is just for a consistent convention.

config/config.go
```diff
@@ -99,11 +99,11 @@ type DatabaseConfig struct {
...
-	URL           string `koanf:"url"`
-	Token         string `koanf:"token"`
-	Org           string `koanf:"org"`
-	Bucket        string `koanf:"bucket"`
-	FlushInterval int    `koanf:"flushinterval"`
+	URL           string        `koanf:"url"`
+	Token         string        `koanf:"token"`
+	Org           string        `koanf:"org"`
+	Bucket        string        `koanf:"bucket"`
+	FlushInterval time.Duration `koanf:"flushinterval"`
```
donch1989, line 106: Didn't notice that we can use `time.Duration` here. It's great. 🚀

cmd/init/definitionupdater/updater.go
jvallesm: I actually prefer standard library / external / Instill (i.e. no separation between alias and no alias), many times adding an alias is just a way to avoid a conflict but in the imports the packages make sense together. However, if this is a convention already discussed, I have no problem changing it.

jvallesm:
Just noticed there's some deadcode in `pkg/service/influx.go`. I'll clean up that bit and re-request the review 🧹

pkg/repository/influx.go
```diff
@@ -0,0 +1,84 @@
...
+)
+
+// InfluxDB reads and writes time series data from InfluxDB.
+type InfluxDB struct {
+	client client.Client
+	api    api.WriteAPI
+}
+
+// NewInfluxDB returns an initialized InfluxDB repository.
+func NewInfluxDB(ctx context.Context) (*InfluxDB, error) {
```
jvallesm, line 24: @donch1989 let me know what you think about this. In [`mgmt-backend`](https://github.com/instill-ai/mgmt-backend/blob/main/pkg/repository/influx.go) InfluxDB is part of the `repository` package, so I think it's good to follow that structure here as well.

cmd/init/definitionupdater/updater.go
donch1989: Yeah, separating the alias is the convention we have discussed before.

pkg/repository/influx.go
donch1989: Can do

PR #509, opened 2024-06-11
feat: render secret value when using dynamic definition

donch1989:
Because

- In the dynamic component definition, we might need to use the secret value to generate the dynamic definition. Therefore, we need to render the secret value for this.

This commit

- Renders the secret value when using dynamic definition.

### DIFF
--- a/pkg/service/convert.go
+++ b/pkg/service/convert.go
@@ -20,6 +20,7 @@ import (
 	"github.com/instill-ai/pipeline-backend/pkg/datamodel"
 	"github.com/instill-ai/pipeline-backend/pkg/logger"
 	"github.com/instill-ai/pipeline-backend/pkg/recipe"
+	"github.com/instill-ai/pipeline-backend/pkg/repository"
 	"github.com/instill-ai/pipeline-backend/pkg/resource"
 
 	componentbase "github.com/instill-ai/component/base"
@@ -50,6 +51,7 @@ type converter struct {
 	redisClient              *redis.Client
 	component                *componentstore.Store
 	aclClient                *acl.ACLClient
+	repository               repository.Repository
 }
 
 // NewService initiates a service instance
@@ -57,6 +59,7 @@ func NewConverter(
 	m mgmtpb.MgmtPrivateServiceClient,
 	rc *redis.Client,
 	acl *acl.ACLClient,
+	r repository.Repository,
 ) Converter {
 	logger, _ := logger.GetZapLogger(context.Background())
 
@@ -65,6 +68,7 @@ func NewConverter(
 		redisClient:              rc,
 		component:                componentstore.Init(logger, nil, nil),
 		aclClient:                acl,
+		repository:               r,
 	}
 }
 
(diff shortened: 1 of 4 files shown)


donch1989:
Hi @chuang8511,
Please check this PR.

pkg/service/pipeline.go
```diff
@@ -631,7 +631,7 @@ func (s *service) preTriggerPipeline(ctx context.Context, isAdmin bool, ns resou
 	for {
 		var nsSecrets []*datamodel.Secret
 		// TODO: should use ctx user uid
-		nsSecrets, _, pt, err = s.repository.ListNamespaceSecrets(ctx, ns.Permalink(), 100, "", filtering.Filter{})
+		nsSecrets, _, pt, err = s.repository.ListNamespaceSecrets(ctx, ns.Permalink(), 100, pt, filtering.Filter{})
 		if err != nil {
```
chuang8511, line 635: NIT: I am not quite sure why we need this part in this PR.
I will trace code more to catch up codebase later.

pkg/service/convert.go
```diff
@@ -141,14 +145,43 @@ func (c *converter) convertResourcePermalinkToName(ctx context.Context, rsc any)
 	return nil
 }
 
-func (c *converter) includeComponentDetail(ctx context.Context, compConfig *componentbase.ComponentConfig, useDynamicDef bool) error {
+func (c *converter) processSetup(ctx context.Context, ownerPermalink string, setup map[string]any) map[string]any {
+	rendered := map[string]any{}
+	for k, v := range setup {
+		switch v := v.(type) {
+		case map[string]any:
+			rendered[k] = c.processSetup(ctx, ownerPermalink, v)
```
chuang8511, line 153: Question:
I checked archetypeap, huggingface, openai.
They seem to be only 1 layer when we setup those components.
I am not sure if it is common thing to have several map in a setup.
If not common, maybe it is too overkilling?

chuang8511 (commented):
Mainly LGTM. Only have a question.
Thanks for helping me do it!
I will directly use this branch to move on my development.

pkg/service/convert.go
```diff
@@ -141,14 +145,43 @@ func (c *converter) convertResourcePermalinkToName(ctx context.Context, rsc any)
...
+		case string:
+			if strings.HasPrefix(v, "${"+recipe.SegSecret+".") && strings.HasSuffix(v, "}") {
```
chuang8511, line 155: Important: I think I will need a else case here.

chuang8511 (commented):
I will add it first for my testing!

pkg/service/pipeline.go
donch1989: It is just a small bug-fix, I put this in the same PR.

pkg/service/convert.go
donch1989: Yeah, you are correct. Fixed.

pkg/service/convert.go
donch1989: Yeah, we don't have this case for now, but adding it doesn't hurt.

PR #513, opened 2024-06-12
fix: improve pipeline usage error messages

jvallesm:
Because

- Error messages on insufficient credit aren't user-fiendly.

This commit
- Removes message extraction from activity error - the type is lost on the wire. Instead, we pass the temporal error generated in the activity, which contains the message.
- Improves activity errors for check / collect activities.


donch1989 (approved):
LGTM

PR #524, opened 2024-06-24
feat: allow user to store recipe in YAML format

donch1989:
Because

- We are going to support pipeline recipes in YAML format. We will return the recipe in YAML format in the response and also allow users to create/update recipes using YAML format.
- Previously, we converted the component definition ID in the recipe into UID when writing it to the database. Now, since we will directly store this raw recipe in the database, we won't do this conversion anymore. In the future, we'll migrate recipes if the component definition ID changes.

This commit

- Exposes the `raw_recipe` field in the pipeline and pipeline_release endpoints, the original `recipe` in JSON format is still workable.
- Removes the component definition ID to UID conversion when reading/writing the recipe from/to the database.

Note

- Regarding the component definition ID to UID conversion, database migration is needed. We'll open another PR to address this soon.
- We only added the most basic integration test for the YAML recipe in this PR. More comprehensive tests will be added after we implement the event trigger pipeline.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -14,7 +14,7 @@ require (
 	github.com/iancoleman/strcase v0.3.0
 	github.com/influxdata/influxdb-client-go/v2 v2.12.3
 	github.com/instill-ai/component v0.20.2-beta
-	github.com/instill-ai/protogen-go v0.3.3-alpha.0.20240531114421-d7be5dd350e5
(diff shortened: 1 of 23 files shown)


donch1989:
Hi @EiffelFly 
Before this PR is merged, you can use this branch to test the YAML recipe.

PR #525, opened 2024-06-25
feat: include imap package

chuang8511:
Because

- we want to add mail component with imap protocol

This commit

- include imap protocol package

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -62,6 +62,7 @@ require (
 	github.com/antchfx/xpath v1.2.4 // indirect
 	github.com/apache/arrow/go/v14 v14.0.2 // indirect
 	github.com/dlclark/regexp2 v1.10.0 // indirect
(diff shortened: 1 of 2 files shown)


donch1989:
@chuang8511 
We should add this to `go.mod` in the component repo. Note that this change won't take effect at all. if you run `go tidy`, this change will disappear.
We should add this in component repo and when the pipeline-backend updates the component package, it will add it automatically.

chuang8511:
> We should add this to go.mod in the component repo.

I also think so. But, yesterday, I tried this but failed to run `go run ./cmd/main` & `go run ./cmd/worker`
After adding these package in pipeline-backend, it succeeded.
I will check the reason soon.

chuang8511:
Ah, my miss. I close this first.
memo
The package will be imported after we update component tag.

PR #529, opened 2024-06-27
feat: refactor pipeline clone endpoints

donch1989:
Because

- When users clone a pipeline, we should provide them with the ability to configure the `description` and `sharing` settings during the cloning process.
- Users might want to clone the pipeline from a specific pipeline release.
- We aim to track the number of times a pipeline has been cloned.

This commit

- Adds `description` and `sharing` fields to the clone endpoints.
- Introduces endpoints for cloning from a pipeline release.
- Includes the `number_of_clones` field in the pipeline statistics.


pkg/errors/errors.go
```diff
@@ -17,4 +17,8 @@ var (
 	ErrInvalidArgument = fmt.Errorf("invalid resource ID")
 	// ErrNotFound is used when a resource doesn't exist.
 	ErrNotFound = fmt.Errorf("not found")
+	// ErrInvalidCloneTarget is used when the pipeline clone target is not
+	// valid. The format should be `<user-id>/<pipeline-id>` or
+	// `<org-id>/<pipeline-id>`.
+	ErrInvalidCloneTarget = fmt.Errorf("invalid target")
```
jvallesm, line 23: 💭 A small note on domain errors, my idea for this package was centralizing the errors that should be communicated between packages. This error is returned by `service` but not intercepted by any package higher up in the execution. 

Do you plan on acting upon this error (e.g. by capturing it in the `middleware` package to return an invalid argument code)? If not , it might still make sense to define this error to avoid repetitio but I'd keep it local to `service` as an unexported error. If yes, I might even follow this strategy and leverage `ErrInvalidArgument`:

```go
package service

// ...

// errInvalidCloneTarget is used when the pipeline clone target is not
// valid. The format should be `<user-id>/<pipeline-id>` or
// `<org-id>/<pipeline-id>`.
var errInvalidCloneTarget = fmt.Errorf("%w: %w", errdomain. ErrInvalidArgument, errInvalidCloneTarget)

// ...

func (s *service) generateCloneTargetNamespace(ctx context.Context, target string) (resource.Namespace, error) {
	targetNamespace, targetID, ok := strings.Cut(target, "/")
	if !ok {
		return resource.Namespace{}, errInvalidCloneTarget
	}
// ...

PR #534, opened 2024-07-01
feat: Intermediate Result Streaming for User Pipelines

tillknuesting:
Because:

We want to get results back when a intermediate step is available not wait until all workflows have concluded

This commit:

Implements gRPC endpoints, service related methods and the query logic to get updates of returned workflows
gRPC endpoints are defined in https://github.com/instill-ai/protobufs/pull/318 INS-4319


config/config.yaml
```diff
@@ -6,7 +6,7 @@ server:
     key:
   edition: local-ce:dev
   usage:
-    enabled: true
+    enabled: false
```
donch1989, line 9: Any reason to turn it off here? I remember you mentioned that the usage client crashes the service under certain circumstances, but I think it should be normal most of the time.

pkg/middleware/middleware.go
```diff
@@ -17,3 +23,207 @@ func AppendCustomHeaderMiddleware(mux *runtime.ServeMux, client pb.PipelinePubli
...
+//		next.ServeHTTP(w, r)
+//	})
+//}
+
+// sseResponseStreamingMiddleware intercepts requests with X-Use-SSE header present
+// and gives back immediately a session token. It continues calling the grpc-gateway
+// endpoint and streams data to the SSE handler using the session ID token.
+func SSEStreamResponseMiddleware(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if r.Header.Get("X-Use-SSE") == "true" {
```
donch1989, line 153: Let's use `Instill-Use-SSE` as our customized header.

pkg/service/pipeline.go
```diff
@@ -962,6 +963,128 @@ func (s *service) triggerPipeline(
...
+		ctx,
+		workflowOptions,
+		"TriggerPipelineWorkflow",
+		&worker.TriggerPipelineWorkflowParam{
+			BatchSize:        len(pipelineData),
+			MemoryStorageKey: memoryKey,
+			SystemVariables: recipe.SystemVariables{
+				PipelineID:          pipelineID,
+				PipelineUID:         pipelineUID,
+				PipelineReleaseID:   pipelineReleaseID,
```
jvallesm, line 1013: This was added recently, can you add the requester UID as well as in [here](https://github.com/instill-ai/pipeline-backend/blob/main/pkg/service/pipeline.go#L1021)? It will determine the credit owner on Cloud. Thanks! 🙇

donch1989 (changes requested):
@tillknuesting Please fix the lint error. Thanks.

https://github.com/instill-ai/pipeline-backend/actions/runs/9779537127

PR #536, opened 2024-07-02
feat: support case-insensitive search for pipelines and components

donch1989:
Because

- When performing fuzzy searches on pipelines and components, case-insensitivity should be allowed.

This commit

- Supports case-insensitive searches for pipelines and components.


jvallesm (approved):
Nice! 🤩 

Can you modify [this test](https://github.com/instill-ai/pipeline-backend/blob/main/integration-test/pipeline/rest-component-definition.js#L87) to e.g. `qTitle="jso"` to cover this change?

jvallesm (approved):

PR #543, opened 2024-07-10
fix: return end-user error on component execution failure

jvallesm:
Because

- By wrapping the temporal application error that comes from `ComponentActivity`, the `service` package failed to extract the error message.

This commit

- Delegate error wrapping to the temporal activities, return activity errors without wrapping.
- In addition to fixing the issue, the error message flow is improved by leveraging the temporal application's `Message` field instead of having an equivalent field within `Details`.

### Before
CleanShot 2024-07-10 at 07 43 10

### After
CleanShot 2024-07-10 at 07 42 20

### DIFF
--- a/pkg/service/pipeline.go
+++ b/pkg/service/pipeline.go
@@ -11,8 +11,6 @@ import (
 	"sync"
 	"time"
 
-	"go.uber.org/zap"
-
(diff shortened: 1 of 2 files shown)


donch1989 (approved):
LGTM

PR #548, opened 2024-07-15
chore: use kebab-case in JSON operator input

jvallesm:
Because

- Migration to `kebab-case` didn't include JQ task in JSON operator.

This commit

- Updates `component` to reference https://github.com/instill-ai/component/pull/219
- Adds a migration to update existing pipelines with `jq` tasks.

### DIFF
--- a/pkg/db/migration/000019.go
+++ b/pkg/db/migration/000019.go
@@ -1,0 +1,132 @@
+package migration
+
+import (
+	"fmt"
+
+	"go.uber.org/zap"
+	"gopkg.in/yaml.v3"
+	"gorm.io/gorm"
+
+	"github.com/instill-ai/pipeline-backend/pkg/datamodel"
+)
+
+const batchSize = 100
+
+type jqInputToKebabCaseConverter struct {
+	db     *gorm.DB
+	logger *zap.Logger
+}
+
+func (c *jqInputToKebabCaseConverter) Migrate() error {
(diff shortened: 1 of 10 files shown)


pkg/db/migration/000019.go
```diff
@@ -0,0 +1,132 @@
+package migration
```
donch1989, line 1: Hi @jvallesm,
I placed the old Go migration script in the convert/xxx package because sometimes we might have structs with the same name in the migration package. Keeping them in different packages helps us avoid this problem.

For example, in different migration steps, we might have structs named `recipe` but with different content.

jvallesm: Yeah I assumed that was the reason so I wanted to check on this. I see the point there, my experience so far with migrations was that I didn't (or barely) need to define structs for the conversion so a dedicated package wasn't required and having the migration code next to the SQL scripts added clarity. In my mind, in case of conflict, we could have structs such as `recipe000019`.

But I have no strong opinion about using the `migration` package vs `convert000019`, and we have code already there so I'll move the Go migration there.

Do you agree that using `migration.go` to centralise the step->migration code makes sense (instead of `cmd/migration/main.go`?

donch1989: > Do you agree that using migration.go to centralise the step->migration code makes sense (instead of cmd/migration/main.go?
Yeah, I think that's a good change.

PR #549, opened 2024-07-16
chore: add kebab-case migration for JSON component

donch1989:
Because

- We missed updating kebab-case for properties inside the JSON component.

This commit

- Adds kebab-case migration for JSON component.


jvallesm:
😅 much simpler approach

PR #551, opened 2024-07-17
fix: parse end-user errors on stream trigger endpoint

jvallesm:
Because

- End-user error messages should only be added when a Temporal error contains a message.
- The stream endpoint wasn't updated with the temporal [error message fix](https://github.com/instill-ai/pipeline-backend/pull/543).

This commit

- Updates the temporal error parsing in the `service` package.


tillknuesting:
lgtm

donch1989 (approved):

PR #556, opened 2024-07-23
feat: add requester information to pipeline metrics

jvallesm:
Because

- In the dashboard, we'll want to display the triggers requested by a namespace (right now we're displaying the triggers on pipelines owned by the namespace). Therefore, we need to add the requester information to the metrics.

This commit

- Adds requester information to pipeline metrics
- Additionally, all the data we might filter / group by is sent as tags. Right now these are fields (not indexed on InfluxDB). When queried, the data is pivoted in order to filter and group.


tillknuesting:
lgtm

PR #559, opened 2024-07-24
feat: use explicit `user_id` and `organization_id` in mgmt request

donch1989:
Because

- Previously, we used a general `name` parameter in request parameters, e.g., `{name=users/*}`, which was ambiguous.

This commit

- Uses explicit `user_id` and `organization_id` in request parameters.
- Removes unnecessary environment variables.


donch1989:
Note: We will release another PR to implement the unified namespace endpoints.

PR #590, opened 2024-08-13
chore(text): migrate text component to new fields

chuang8511:
Because

- we update text component

This commit

- migrate fields to current design


chuang8511:
I will rebase it in sprint 43

PR #592, opened 2024-08-14
feat(vdp): pipeline run logging

joremysh:
Because

This commit


donch1989:
@joremysh I think we can close this one, right?

PR #593, opened 2024-08-15
feat(minio): added minio connection

joremysh:
Because

- backend needs to store trigger input & output & files

This commit

- added minio connection


joremysh:
lint passed on local
image

donch1989 (approved):
LGTM

PR #595, opened 2024-08-15
feat(run): run logging data model & repo func

joremysh:
Because

- we need to store run logging data

This commit

- added data model, migration SQL, repo functions

tested locally with below and passed

``` bash
go test -v ./... -tags dbtest -run TestRepository_UpsertPipelineRun
```


donch1989 (approved):
LGTM

PR #598, opened 2024-08-16
feat(run): uploaded pipeline run logging output and component inputs

joremysh:
Because

- run logging needs to store input & output for each component

This commit

- uploaded pipeline run logging output and component inputs

ps: after testing on main branch, it seems TriggerUserPipelineWithStream does not work. It will TriggerPipelineWorkflow 1000 times in 1 sec.


joremysh:
rebased the changes here https://github.com/instill-ai/pipeline-backend/pull/602

PR #600, opened 2024-08-20
chore(document): update the key

chuang8511:
Because

- we update the document operator field

This commit

- migrate the field name

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -26,7 +26,7 @@ database:
   host: pg-sql
   port: 5432
   name: pipeline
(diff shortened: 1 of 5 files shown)


chuang8511:
Will do migration setting when I have to fix conflict from other branch.

chuang8511:
@donch1989 
Here.
https://github.com/instill-ai/component/pull/291#pullrequestreview-2248301841

Because there could be multiple migration in a Sprint, how about we set up all config before the release in the separate PR?

PR #606, opened 2024-08-23
feat(run): add dataSpecification in pipeline run logging response for…

joremysh:
… FE to specify schema

Because

- FE need dataSpecification to format input & output

This commit

- convert recipe to dataSpecification and add to response


donch1989 (approved):
LGTM

PR #613, opened 2024-08-26
feat: add supports for Data properties

donch1989:
Because

- We introduced data structures for unstructured data, and we want users to easily access the data and perform format conversions.

This commit

- Refactors the reference system.
- Adds support for data properties and format conversions. For example, users can get the image width via `${some-component.image.width}` or convert a PDF to markdown via `${variable.document.markdown}`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,8 +20,18 @@ dev:							## Run dev container
 	@docker inspect --type container ${SERVICE_NAME} >/dev/null 2>&1 && echo "A container named ${SERVICE_NAME} is already running." || \
 		echo "Run dev container ${SERVICE_NAME}. To stop it, run \"make stop\"."
 	@docker run -d --rm \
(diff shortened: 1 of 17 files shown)


Makefile
```diff
@@ -20,8 +20,18 @@ dev:							## Run dev container
 	@docker inspect --type container ${SERVICE_NAME} >/dev/null 2>&1 && echo "A container named ${SERVICE_NAME} is already running." || \
 		echo "Run dev container ${SERVICE_NAME}. To stop it, run \"make stop\"."
 	@docker run -d --rm \
-		-v $(PWD):/${SERVICE_NAME} \
 		-p ${SERVICE_PORT}:${SERVICE_PORT} \
+		-v $(PWD)/../go.work:/go.work \
```
jvallesm, line 24: @donch1989 I think this wasn't intentional

PR #620, opened 2024-08-28
chore: fix iterator memory and run-logging bug

donch1989:
Because

- The run-logging API errors when there’s an iterator in the log.
- The iterator workflow uses incorrect workflow memory.

This commit

- Fixes the iterator memory and run-logging bugs.


pkg/service/pipeline.go
```diff
@@ -1811,6 +1811,10 @@ func (s *service) ListPipelineRuns(ctx context.Context, req *pipelinepb.ListPipe
 					return nil, fmt.Errorf("failed to load recipe metadata. pipeline UID: %s recipe reference ID: %s", run.PipelineUID.String(), run.RecipeSnapshot[0].Name)
 				}
 
+				err = s.converter.IncludeDetailInRecipe(ctx, "", dbRecipe, false)
```
donch1989, line 1814: Hi @joremysh 
I found a run-logging bug when a pipeline includes an iterator. I've fixed it in this commit.

PR #626, opened 2024-08-29
chore(document): take out the unnecessary dependency

chuang8511:
Because

- pdf2md is temp solution

This commit

- take out the dependency for pdf2md


chuang8511:
cc @donch1989

PR #630, opened 2024-09-01
chore(service): remove unnecessary condition check

septemhill:
~~Return error if temporal client is `nil`, if return `nil` back to user, if would be treat no error happened and ran successful~~

Edit: after review again, the `temporalClient` would be assigned in main function, so it would never be `nil`

### DIFF
--- a/pkg/service/convert.go
+++ b/pkg/service/convert.go
@@ -978,7 +978,7 @@ func (c *converter) GeneratePipelineDataSpec(variables map[string]*datamodel.Var
 			})
 		}
 
-		if err != nil {
+		if m == nil || err != nil {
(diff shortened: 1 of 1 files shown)


jvallesm (approved):
> Edit: after review again, the temporalClient would be assigned in main function, so it would never be nil

Indeed, we follow the convention that `service`'s methods trust the receiver to be initialised, so this check isn't aligned with that.

Thanks for the contribution!

septemhill:
@jvallesm  Ah, sorry.

I think I made something wrong. I figure out there is CI error, then I rebase the latest code, and this PR be closed 
I'll create another one for this.

PR #631, opened 2024-09-02
chore(service): remove unnesssary condition check

septemhill:
after review again, the temporalClient would be assigned in main function, so it would never be nil

### DIFF
--- a/pkg/service/pipeline.go
+++ b/pkg/service/pipeline.go
@@ -301,10 +301,14 @@ func (s *service) GetPipelineByUIDAdmin(ctx context.Context, uid uuid.UUID, view
 }
 
 func (s *service) setSchedulePipeline(ctx context.Context, ns resource.Namespace, pipelineID, pipelineReleaseID string, pipelineUID, releaseUID uuid.UUID, recipe *datamodel.Recipe) error {
-
+	// TODO This check could be removed, as the receiver should be initialized
+	// at this point. However, some tests depend on it, so we would need to
+	// either mock this interface or (better) communicate with Temporal through
+	// our own interface.
 	if s.temporalClient == nil {
 		return nil
 	}
+
 	crons := []string{}
 	if recipe != nil && recipe.On != nil && recipe.On.Schedule != nil {
 		for _, v := range recipe.On.Schedule {


jvallesm (commented):
Hi @septemhill can you rebase your branch? The lint CI step was broken, the fix was just merged into `main`

septemhill:
@jvallesm sure, done

jvallesm:
The `service` package unit tests are failing because of your change. This `nil` check allows tests to bypass the Temporal dependencies in some tests. To fix it, you'd need to inject a `temporalclient.Client` mock instead of an empty interface on `service.Service` creation, and mock the affected methods.

An alternative we've used in some repos is defining a `Scheduler` interface that abstracts all interaction with the `temporal` packages, then injecting a mock of this interface, which tends to have a smaller API and reduces the external dependencies in the `service` package.

In any case, either alternative is a prerequisite to remove the `nil` check, otherwise we'd be reducing our coverage.

septemhill:
> In any case, either alternative is a prerequisite to remove the `nil` check, otherwise we'd be reducing our coverage.

After reviewing the code, I figured out the code depends on the `Client` interface of `temporal` library directly.
Usually, I'd like to maintain my own interface for service layer, it would more easy to make mock (use codegen tool, e.g: mockery, gomock, etc)

Here, it became a tricky problem. So 3 ways we have: 

- DO NOT make this change. Instead, add comment here to remind people the section for mock testing only, kind of workaround. 
- Create `Client` mock manually. (worse case, IMO)
- Create our own interface to use  temporal package, rather than use it directly (best case, IMO)

@jvallesm  Any idea ?

jvallesm:
@septemhill your assesment is correct. Among the options you provide, 3️⃣ is the cleaner and more long-term solution, but it's also the one that implies the most changes in the production code. Moreover, these are a package and a function that are fundamental to the product. The approach we've been following is introducing this very approach (abstract the interaction with Temporal through an interface) in new features, so we don't feed the old pattern and we don't add work to our future selves when we tackle refactoring this package. I'm sorry I can provide examples about how the new approach looks, they were introduced in 2 private repositories recently, but it's likely that new features in the public repos will soon replicate that pattern.

As mentioned in another PR, we've been moving fast and pushing quite some code this year to bring many functionalities to Instill VDP. We estimate we'll be able to slow a bit down in Q4 and dedicate more time to improve the code quality, coverage and performance.

Among the other 2 options, I'm fine with either. You may try mocking the `Client` interface with a no-op implementation (we use `minimock` and declare the mocks in `pkg/mock`). But, given the amount of production code we're talking about here (3 lines fixing one redundant check), I think it's valid to drop change and wait for the refactorisation of this method. In this case a TODO is useful to remind us why the check is there, something in the lines of:

```go
	// TODO This check could be removed, as the receiver should be initialized
	// at this point. However, some tests depend on it, so we would need to
	// either mock this interface or (better) communicate with Temporal through
	// our own interface.
	if s.temporalClient == nil {
		return nil
	}
```

septemhill:
@jvallesm, 

Agree. If we'd like to refactor in the future, then add comment is better choice for current situation.

jvallesm (approved):

PR #634, opened 2024-09-03
chore(all): improve code quality and readability lightly

septemhill:
- Reduce memory duplicate (NewBuffer -> NewBufferString)
- Remove variable reassignment
- Add `TODO` comments for things should be take care

### DIFF
--- a/cmd/init/presetdownloader/downloader.go
+++ b/cmd/init/presetdownloader/downloader.go
@@ -111,7 +111,7 @@ func DownloadPresetPipelines(ctx context.Context, repo repository.Repository) er
 	}
 
 	clientConn, err := grpc.NewClient(fmt.Sprintf("%s:%d", config.Config.InstillCloud.Host, config.Config.InstillCloud.Port),
-		grpc.WithTransportCredentials(credentials.NewTLS((&tls.Config{}))),
+		grpc.WithTransportCredentials(credentials.NewTLS((&tls.Config{MinVersion: tls.VersionTLS12}))),
 		grpc.WithDefaultCallOptions(grpc.MaxCallRecvMsgSize(constant.MaxPayloadSize), grpc.MaxCallSendMsgSize(constant.MaxPayloadSize)))
 	if err != nil {
 		// Skip the download process if Instill Cloud is unreachable.
--- a/cmd/main/main.go
+++ b/cmd/main/main.go
@@ -218,6 +218,7 @@ func main() {
 	if config.Config.Server.HTTPS.Cert != "" && config.Config.Server.HTTPS.Key != "" {
 		tlsConfig = &tls.Config{
 			ClientAuth: tls.RequireAndVerifyClientCert,
+			MinVersion: tls.VersionTLS12,
 		}
 		creds, err = credentials.NewServerTLSFromFile(config.Config.Server.HTTPS.Cert, config.Config.Server.HTTPS.Key)
 		if err != nil {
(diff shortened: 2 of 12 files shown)


jvallesm:
Hi @septemhill, thanks for the contribution. We're finishing a release so I'll have more time to review these PRs by the end of the week.

About this,
> Add TODO comments for things should be take care

we've been moving fast in order to ship features and there's plenty of room for improvement in the codebase. We realised we were writing more and more TODOs but we didn't know where we were going to tackle, so we're avoiding now to put such comments in the codebase if they don't have an action plan (e.g. a ticket / GitHub issue / release version attached). Therefore, I encourage you to open an issue with your TODO comments (I think one grouping all these improvements is fine), which can be prioritised and tracked by the team or picked up by a community member.

septemhill:
Hi @jvallesm 

I did think about open an issue, but this repo didn't provide the tab

So I'll remove the TODO section later

jvallesm:
We've centralised issues in https://github.com/instill-ai/instill-core/issues, as some of them are cross-repository.

septemhill:
@jvallesm, just move those `TODO` the issue.

cmd/main/main.go
```diff
@@ -357,15 +358,17 @@ func main() {
 	}
 
 	privateHTTPServer := &http.Server{
-		Addr:      fmt.Sprintf(":%v", config.Config.Server.PrivatePort),
-		Handler:   grpcHandlerFunc(privateGrpcS, privateServeMux),
-		TLSConfig: tlsConfig,
+		Addr:              fmt.Sprintf(":%v", config.Config.Server.PrivatePort),
+		Handler:           grpcHandlerFunc(privateGrpcS, privateServeMux),
+		TLSConfig:         tlsConfig,
+		ReadHeaderTimeout: 10 * time.Millisecond,
```
jvallesm, line 364: This is a bit aggressive (although for headers should be fine). What did you base your choice on?

cmd/init/presetdownloader/downloader.go
```diff
@@ -111,7 +111,7 @@ func DownloadPresetPipelines(ctx context.Context, repo repository.Repository) er
 	}
 
 	clientConn, err := grpc.NewClient(fmt.Sprintf("%s:%d", config.Config.InstillCloud.Host, config.Config.InstillCloud.Port),
-		grpc.WithTransportCredentials(credentials.NewTLS((&tls.Config{}))),
+		grpc.WithTransportCredentials(credentials.NewTLS((&tls.Config{MinVersion: tls.VersionTLS12}))),
```
jvallesm, line 114: Nothing against being explicit about the minimum version but this is also the default value, so I'd be fine with `&tls.Config{}`, too.

jvallesm (approved):
A couple of minor comments, not requiring changes. The CI step for integration tests only failed on teardown, so coverage is correct 👍

cmd/init/presetdownloader/downloader.go
septemhill: The default min version would have security issue, that's why I need to assign a specific version.

cmd/main/main.go
septemhill: Security issue (https://cwe.mitre.org/data/definitions/400.html)
If we don't setup the header timeout, what if we have million or billion request come in with malice, it would consume all of our service resource.

jvallesm: Yeah, I meant more the actual value (10ms vs, for instance, 100ms). I agree it's better to cap the header timeout.

cmd/init/presetdownloader/downloader.go
jvallesm: But you assigned the same version as the [default min version](https://pkg.go.dev/crypto/tls#Config) 🤔 
```
	// By default, TLS 1.2 is currently used as the minimum. TLS 1.0 is the
	// minimum supported by this package.
```

cmd/main/main.go
septemhill: Ah, I missed. The HTTP header parser usually wouldn't take over 1 ms (for 1KB size). So 5-10 ms is enough. Unless our HTTP header over 1KB, maybe we need to set more longer.

cmd/init/presetdownloader/downloader.go
septemhill: You're right. I've checked the Go 1.22.5 (the go.mod using this version) and 1.23, doc did mention the same thing. 
It seems `gosec`'s issue. If you don't set the MinVersion explicitly with specific version, it would keep warning you.
You want keep this change ? Or I can rollback

jvallesm: It's fine as it is, being explicit about these things never hurts 👍

PR #638, opened 2024-09-05
feat(integration): reference connections in pipelines

jvallesm:
Because

- After creating a connection, users need to be able to reference it from a pipeline.

This commit

- (fix) Makes connection ID mutable
- Resolves connection references on pipelines (e.g. `setup: ${connection.my-cohere-conn}`).


pkg/worker/io.go
```diff
@@ -24,7 +24,6 @@ func NewSetupReader(wfm memory.WorkflowMemory, compID string, conditionMap map[i
 }
 
 func (i *setupReader) Read(ctx context.Context) (setups []*structpb.Struct, err error) {
-
 	for idx := range len(i.conditionMap) {
 		setupTemplate, err := i.wfm.GetComponentData(ctx, i.conditionMap[idx], i.compID, memory.ComponentDataSetup)
```
jvallesm, line 28: Here (and in other places), when I logged the value I got a `BADKEY` string in the output:

```
2024/09/05 14:19:47 INFO +++++ SETUP TEMPLATE !BADKEY="string_value:\"${connection.my-openai}\""
```

However, this doesn't mean the value is not there and the pipeline trigger works as expected. @donch1989 have you faced this when working with the workflow memory?

donch1989: Hi @jvallesm,
I just tried some pipelines, and the setup rendering looks good. Could you share how you're logging this value? I can try again and check from my side.

jvallesm:
```go
func (i *setupReader) Read(ctx context.Context) (setups []*structpb.Struct, err error) {
	for idx := range len(i.conditionMap) {
		setupTemplate, err := i.wfm.GetComponentData(ctx, i.conditionMap[idx], i.compID, memory.ComponentDataSetup)
		if err != nil {
			return nil, err
		}

		val, _ := setupTemplate.ToStructValue()
		slog.Info("+++++ SETUP TEMPLATE", val) // THIS shows BADKEY

		setupVal, err := recipe.Render(ctx, setupTemplate, i.conditionMap[idx], i.wfm, false)
		if err != nil {
			return nil, err
		}
		setup, err := setupVal.ToStructValue()
		if err != nil {
			return nil, err
		}
		setups = append(setups, setup.GetStructValue())
	}

	return setups, nil
}
```

donch1989: I think the issue is caused by a missing key in the slog. If you use `slog.Info("+++++ SETUP TEMPLATE", "setup", val)`, it should work fine.

jvallesm: Oh 🤦 I went quick on the logging and didn't introduce the extra args as key-value. I thought it was the memory store rather than the log itself. Thanks for the eye!

donch1989:
Another issue is that we can't use `mgmtPrivateClient` during the integration test teardown.

pkg/service/secret.go
```diff
@@ -111,9 +111,20 @@ func (s *service) DeleteNamespaceSecretByID(ctx context.Context, ns resource.Nam
...
 
-	for k, v := range setup {
+	switch s := setup.(type) {
+	case map[string]any:
+		setupMap = s
+	case string:
+		// Setup should contain a reference to a connection.
+		return nil
+	default:
+		return fmt.Errorf("invalid type for setup field")
```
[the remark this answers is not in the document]
jvallesm: Yep I saw that debugging the integration tests, added the `nil` condition [here](https://github.com/instill-ai/pipeline-backend/pull/638/commits/22c0edf71418420953bc51b2085c97da104a9927#diff-a4e8445d22c3d9998832a49a2063ffc12ead15fdc0377928188b2bef78ceaf82R137) and [here](https://github.com/instill-ai/pipeline-backend/pull/638/commits/22c0edf71418420953bc51b2085c97da104a9927#diff-73f8f8962dc6fc05cf664d9b4d970e57a16f81dd10da742f3e7eba1befb59fc1R120).

jvallesm:
> Another issue is that we can't use `mgmtPrivateClient` during the integration test teardown. Since the CE version only allows a single user, we don't need to filter the `namespace_uid` during the teardown.

Indeed, I'd started by deleting all the connections for that user and I didn't have another way to access the UID (but to hardcode it, which I didn't like). But later I thought that the environment isn't created and torn down with the tests and that it might erase some pipelines in the dev environment, which would be inconvenient, so I added the ID prefix. With the prefix, we don't need the user anymore.

jvallesm:
@donch1989 I removed the gRPC call to `mgmt-backend` in the tests and modified the database host to `localhost` (on the CI the tests aren't run from the pipeline container itself). I just added a condition to read the DB host from an environment variable and default to `localhost`, note that now if we run the tests from within the pipeline container we'll need to add `DB_HOST=pg-sql`.

PR #660, opened 2024-09-20
chore(component): update dependency

chuang8511:
Because

- we update component

This commit

- update component


jvallesm (changes requested):
Please check the CI, the build is broken. I think the only thing you need to do is update the `protogen-go` package. I added the `protoc-gen-validator` plugin yesterday and had to introduce some fixes afterwards due to some name conflicts in the generated code.

chuang8511:
@jvallesm 
Thanks for noticing!

jvallesm (approved):

PR #661, opened 2024-09-20
feat(integration): support OAuth connections

jvallesm:
Because

- Some components support OAuth 2.0 integrations to fill their setup.

This commit

- Allows `METHOD_OAUTH` in connections and OAuth-specific fields.
- Returns single setup for integrations.


jvallesm:
ℹ️ `golangci-lint` action is failing because we keep using a deprecated proto field. This will be kept for one release to let the frontend switch from the old field to the new one.

PR #670, opened 2024-09-23
feat(component): implement instill app features

chuang8511:
Because

- instill app is only exposed to ee version

This commit

- exclude instill app when it is ce version
- initialise Instill App
- fix the bug in Instill App when end to end test
- chore the exported and non-exported functions

Note
- We should extend the OneOf function for output in the other PR in near future.

### DIFF
--- a/pkg/service/component_definition.go
+++ b/pkg/service/component_definition.go
@@ -11,6 +11,7 @@ import (
 	"google.golang.org/protobuf/proto"
 	"google.golang.org/protobuf/reflect/protoreflect"
 
+	"github.com/instill-ai/pipeline-backend/config"
 	"github.com/instill-ai/pipeline-backend/pkg/recipe"
 	"github.com/instill-ai/pipeline-backend/pkg/repository"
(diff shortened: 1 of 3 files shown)


pkg/service/component_definition.go
```diff
@@ -222,6 +223,14 @@ func (s *service) ListComponentDefinitions(ctx context.Context, req *pb.ListComp
 		if req.GetView() != pb.ComponentDefinition_VIEW_FULL {
 			def.Spec = nil
 		}
+
+		// Temporary solution to filter out the instill-app component for local-ce:dev edition
+		// In the future, when it is a confirmed requirement, we can add a new field in the
+		// component definition to mark it as hidden.
```
jvallesm, line 229: 💭 In the future the non-CE definitions shouldn't be in the CE repositories and Cloud should extend the CE component store to initialise the private components.

chuang8511: Yes, we think so. Thanks for pointing out. I will add it in the comment as well.

jvallesm (approved):

PR #673, opened 2024-09-25
fix(migration): fix migration bug

chuang8511:
Because

- there is a bug in migration file

This commit

- fix the bug

### DIFF
--- a/pkg/db/migration/convert/convert000029/convert.go
+++ b/pkg/db/migration/convert/convert000029/convert.go
@@ -32,6 +32,10 @@ func (c *ConvertToArtifactType) migratePipeline() error {
 			isRecipeUpdated := false
 			l := c.Logger.With(zap.String("pipelineUID", p.UID.String()))
 
+			if p.Recipe == nil {
+				continue
+			}
+
 			for id, comp := range p.Recipe.Component {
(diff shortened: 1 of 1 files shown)


jvallesm (approved):

pkg/db/migration/convert/convert000029/convert.go
```diff
@@ -45,7 +49,7 @@ func (c *ConvertToArtifactType) migratePipeline() error {
 			}
 
 			if isRecipeUpdated {
-				recipeYAML, err := yaml.Marshal(p.Recipe)
+				recipeYAML, err := yaml.Marshal(p.RecipeYAML)
```
donch1989, line 52: Hi @chuang8511,

1. Will `yaml.Marshal(p.RecipeYAML)` preserve all the comments and layout from the original YAML?
2. I have doubts about this working as expected since you're migrating `p.Recipe`, but you're saving `p.RecipeYAML`.

donch1989 (changes requested):

pkg/db/migration/convert/convert000029/convert.go
chuang8511: Sorry, I did not notice this part. I will check it later.

PR #675, opened 2024-09-26
feat(document): add dependency

chuang8511:
Because

- some pdf cannot be processed by pdfplumber

This commit

- add the dependency of the tool to repair PDF


chuang8511:
We can directly use libreoffice

PR #680, opened 2024-09-30
feat(universalai): add instill credit function for universal ai component

jvallesm:
This is a port of https://github.com/instill-ai/component/pull/362

Because

- we will support universal ai to execute vendors' ai models

This commit

- import the vendors' key to component and execution

### DIFF
--- a/pkg/component/ai/universalai/v0/config/tasks.json
+++ b/pkg/component/ai/universalai/v0/config/tasks.json
@@ -15,60 +15,6 @@
           "instillShortDescription": "Input data",
           "type": "object",
           "properties": {
-            "model": {
-              "description": "The model to be used. Now, it only supports OpenAI model, and will support more models in the future.",
-              "instillShortDescription": "The model to be used.",
-              "instillAcceptFormats": [
-                "string"
-              ],
-              "enum": [
-                "o1-preview",
-                "o1-mini",
-                "gpt-4o-mini",
-                "gpt-4o",
-                "gpt-4o-2024-05-13",
-                "gpt-4o-2024-08-06",
-                "gpt-4-turbo",
-                "gpt-4-turbo-2024-04-09",
-                "gpt-4-0125-preview",
-                "gpt-4-turbo-preview",
-                "gpt-4-1106-preview",
-                "gpt-4-vision-preview",
-                "gpt-4",
-                "gpt-4-0314",
-                "gpt-4-0613",
-                "gpt-4-32k",
-                "gpt-4-32k-0314",
-                "gpt-4-32k-0613",
-                "gpt-3.5-turbo",
-                "gpt-3.5-turbo-16k",
-                "gpt-3.5-turbo-0301",
-                "gpt-3.5-turbo-0613",
-                "gpt-3.5-turbo-1106",
-                "gpt-3.5-turbo-0125",
-                "gpt-3.5-turbo-16k-0613"
-              ],
-              "instillCredentialMap": {
-                "values": [
-                  "o1-preview",
-                  "o1-mini",
-                  "gpt-4o",
-                  "gpt-4o-2024-08-06",
-                  "gpt-4-turbo",
-                  "gpt-4-vision-preview",
-                  "gpt-4",
-                  "gpt-4-32k",
-                  "gpt-3.5-turbo",
-                  "gpt-4o-mini"
-                ],
-                "targets": [
-                  "setup.api-key"
-                ]
-              },
-              "instillUIOrder": 0,
-              "title": "Model Name",
-              "type": "string"
-            },
             "messages": {
               "title": "Chat Messages",
               "type": "array",
@@ -203,12 +149,11 @@
                   "role"
                 ]
               },
-              "instillUIOrder": 1,
+              "instillUIOrder": 0,
               "description": "List of chat messages"
             }
           },
           "required": [
-            "model",
             "messages"
           ],
           "instillUIOrder": 0
--- a/pkg/component/ai/universalai/v0/main.go
+++ b/pkg/component/ai/universalai/v0/main.go
@@ -36,11 +36,12 @@ var (
 type component struct {
 	base.Component
 
-	instillAPIKey string
+	instillAPIKey map[string]string
 }
 
 type execution struct {
 	base.ComponentExecution
+
 	usesInstillCredentials bool
 	execute                func(*structpb.Struct, *base.Job, context.Context) (*structpb.Struct, error)
 }
@@ -59,7 +60,12 @@ func Init(bc base.Component) *component {
 }
 
 func (c *component) CreateExecution(x base.ComponentExecution) (base.IExecution, error) {
-	resolvedSetup, resolved, err := c.resolveSetup(x.Setup)
+
+	model := getModel(x.GetSetup())
+	vendor := ModelVendorMap[model]
+
+	resolvedSetup, resolved, err := c.resolveSetup(vendor, x.Setup)
+
 	if err != nil {
 		return nil, err
 	}
@@ -92,15 +98,18 @@ func (e *execution) UsesInstillCredentials() bool {
 // WithInstillCredentials loads Instill credentials into the component, which
 // can be used to configure it with globally defined parameters instead of with
 // user-defined credential values.
-func (c *component) WithInstillCredentials(s map[string]any) *component {
-	c.instillAPIKey = base.ReadFromGlobalConfig(cfgAPIKey, s)
+func (c *component) WithInstillCredentials(vendor string, s map[string]any) *component {
+	if c.instillAPIKey == nil {
+		c.instillAPIKey = make(map[string]string)
+	}
+	c.instillAPIKey[vendor] = base.ReadFromGlobalConfig(cfgAPIKey, s)
 	return c
 }
 
 // resolveSetup checks whether the component is configured to use the Instill
 // credentials injected during initialization and, if so, returns a new setup
 // with the secret credential values.
-func (c *component) resolveSetup(setup *structpb.Struct) (*structpb.Struct, bool, error) {
+func (c *component) resolveSetup(vendor string, setup *structpb.Struct) (*structpb.Struct, bool, error) {
 	if setup == nil || setup.Fields == nil {
 		setup = &structpb.Struct{Fields: map[string]*structpb.Value{}}
 	}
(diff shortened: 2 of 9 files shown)


pkg/component/base/execution_wrapper.go
```diff
@@ -108,6 +108,14 @@ func (e *ExecutionWrapper) Execute(ctx context.Context, jobs []*Job) (err error)
 	for batchIdx, job := range wrappedJobs {
 		outputs[batchIdx] = job.Output.(*outputWriter).GetOutput()
 	}
+
+	// Universal AI component set usesInstillCredentials when it executes, which is different from other AI components.
+	h, err = newUH(e)
```
jvallesm, line 113: ❓ Why is this `h` different than the one declared above?

chuang8511: Ah, maybe the comment is misleading.

Because Universal AI component set usesInstillCredentials when it executes, I have to re-initialise the h after the job executes.

So, they are same, and I just re-initialise it.
I will change the comment

jvallesm: I don't understand why this needs to be computed during `Execute` instead of `CreateExecution`. Can you expand on this?

chuang8511: > why this needs to be computed during Execute instead of CreateExecution.

Because it will need to know which vendor to be called when the input is inserted, which is in the process of Execute.

> Having a component that changes this under the hood and forcing the interface it implements to adapt to that fact feels like an antipattern.

I do not get this point.
Do you mean that it will be better pattern to initialise another handler?
Or, maybe we should differentiate the handler to calculate the input token and the output cost.
Now, it is 1 handler. Check is more like before the Execute.
But, Collect is more like after the Execute because the output usually comes after the execution.
Is this point you are talking about?

```
// UsageHandler allows the component execution wrapper to add checks and
// collect usage metrics around a component execution.
type UsageHandler interface {
	Check(ctx context.Context, inputs []*structpb.Struct) error
	Collect(ctx context.Context, inputs, outputs []*structpb.Struct) error
}
```

donch1989: @chuang8511 
What if we put the model name into setup?

donch1989: The issue arises because we can only resolve secrets in `setup`, not in `input`.
Maybe we should reconsider which data belongs to `setup` and which data belongs to `input`.

chuang8511: > The issue arises because we can only resolve secrets in setup, not in input.

Why do we have this limitation?

> Maybe we should reconsider which data belongs to setup and which data belongs to input.

Yes. I think so. The model is a bit weird for me to set it in setup.
But, it seems a good solution too.

donch1989: Although it may not be consistent with other components, placing it in `setup` is still valid. The original purpose of having separate `setup` and `input` sections is to distinguish between 1. how we configure or connect the component and 2. the data fed into it. We can consider `model_name` as a configuration rather than part of the input data.

chuang8511: > But recreating the usage handler because one implementation changes it during execution is obscure, it can lead to hard-to-debug issues.

It is a good opinion.

> I'd vote for keeping this in the setup but I'm open for alternatives as long as the interface is clear and dictates the implementations, not the other way around.

> Although it may not be consistent with other components, placing it in setup is still valid.

Ok, I will modify it. Thanks for feedback.

chuang8511: I move it to setup.

jvallesm (commented):
OK, now (reading the latest changes) I see the problem with having the model in the setup. Model is tied to Task, and Task is an input param, so having this in the setup is far from ideal. It produces the same obscurity I didn't like in modifying the output of `UsesInstillCredentials` from within an execution. In this case, we'll wonder "why this component in particular has the model in the setup? doesn't this determine the task than will be later set in the input?". Setup is not for this and, now, with the Integration feature, I don't see it as a clean solution (I imagine creating a connection and needing to specify the model).

As mentioned yesterday, UniversalAI component challenges the assumptions based on which the `IExecution` interface was modified to accommodate the usage handler / credit consumption feature. Before, we could know whether a component execution would be using a global key at the `CreateExecution` moment because it depended solely on the setup and on the initialisation of the component. Now, it will depend on the execution input. So we can pass the input to `UsesInstillCredentials` to check if that execution uses credentials / should consume credit. This has another trap: we're assuming that all the jobs in the input have the same model, which isn't enforced anywhere in the code. Right now we have a usage handler per execution, not per input in the batch of an execution. We'll face this problem with either solution at this point (recreating the usage handler after execution or passing the input).

All this doesn't need to be taken into account for this PR, I'm aware we're at the end of a sprint and compromises need to be taken. But I'd like for us to consider these issues and refactor the usage handler pattern as we move forward with the UniversalAI, I feel the current solutions just hack the component to make it behave the way we'd like to, adding debt that will be hard to clean up in the future.

chuang8511:
>  It produces the same obscurity I didn't like in modifying the output of UsesInstillCredentials from within an execution.

I do not get this. Previously, it was. But, now, we do not modify output of UsesInstillCredentials from within an execution. right? Do I miss to read some points?

> In this case, we'll wonder "why this component in particular has the model in the setup? doesn't this determine the task than will be later set in the input?". Setup is not for this and, now, with the Integration feature, I don't see it as a clean solution (I imagine creating a connection and needing to specify the model).

To clarify, I thought it is a trade-off point in the yesterday discussion.
We do not want one implementation to change the interface design, so we accept model to put in setup.
I do not get a new POV about why it is a clean solution.

And, about "(I imagine creating a connection and needing to specify the model).", we may need time to discuss with product/UX team how it will be. I think the current "unclean" code design is mainly from how we imagine the product differently.


> This has another trap: we're assuming that all the jobs in the input have the same model, which isn't enforced anywhere in the code. Right now we have a usage handler per execution, not per input in the batch of an execution. We'll face this problem with either solution at this point (recreating the usage handler after execution or passing the input).

I think it is a good point to foresee the problem. I will list it in the ticket for the refactor after we finish this.

pkg/component/ai/universalai/v0/config/tasks.json
```diff
@@ -15,60 +15,6 @@
           "instillShortDescription": "Input data",
           "type": "object",
           "properties": {
-            "model": {
```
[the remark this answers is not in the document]
chuang8511: I will modify it! Thank you for pointing out!

pkg/component/ai/universalai/v0/main.go
```diff
@@ -111,10 +120,42 @@ func (c *component) resolveSetup(setup *structpb.Struct) (*structpb.Struct, bool
...
-	setup.GetFields()[cfgAPIKey] = structpb.NewStringValue(c.instillAPIKey)
+	setup.GetFields()[cfgAPIKey] = structpb.NewStringValue(c.instillAPIKey[vendor])
 	return setup, true, nil
 }
+
+func getModel(setup *structpb.Struct) string {
+	return setup.GetFields()["model"].GetStringValue()
+}
+
+var ModelVendorMap = map[string]string{
```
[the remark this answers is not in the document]
chuang8511: Good catch. Thank you. I will modify it.

BTW, I am curious how you decide if a var or function is exposed or not.
Is that as simple as that whether it needs to be known by other packages?

pkg/component/store/store.go
```diff
@@ -123,11 +124,12 @@ func Init(
...
 		}
-		// {
-		// 	conn := universalai.Init(baseComp)
-		// 	conn = conn.WithInstillCredentials(secrets[conn.GetDefinitionID()])
-		// 	compStore.Import(conn)
-		// }
+		{
+			conn := universalai.Init(baseComp)
+			// Please apply more keys when we add more vendors
+			conn = conn.WithInstillCredentials("openai", secrets["openai"])
```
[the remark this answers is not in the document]
chuang8511: > The signature of WithInstillCredentials won't work in the future when we want to pass the secrets of several vendors. 

Why won't it work in the future?
I overwrote [the function](https://github.com/instill-ai/pipeline-backend/pull/680/files#diff-ff20f7796e9b08cfaf588f5acb05c3445db65093c50b61b0dd6278e428ecc34dR101).

Or, do you mean, ideally, we shouldn't overwrite any interface function?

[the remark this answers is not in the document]
jvallesm: > Or, do you mean, ideally, we shouldn't overwrite any interface function?

No, not that. Actually this function isn't part of any interface. In my comment I was wondering if it made sense to pass `secrets` as a whole in this case, or another strategy (I don't like passing unnecessary secrets to the component). If you have several vendors, I guess you can call `conn.WithInstillCredentials` several times, which is a possible solution.

jvallesm:
> I do not get this. Previously, it was. But, now, we do not modify output of UsesInstillCredentials from within an execution. right? Do I miss to read some points?

It was an example of what I critiziced previously. Now you fixed that but by putting Model in the setup (as I recommended) we just made another part unclear. So in my comment I meant that both solutions work around a problem that lies somewhere else: the `UsageHandler` interface was designed to be reusable during the whole lifespan of a `ComponentExecution`, but UniversalAI component breaks this assumption and we need to adapt.

> To clarify, I thought it is a trade-off point in the yesterday discussion

Exactly, let me stress this from my previous comment: at this point I'm fine with re-creating the usage handler or putting the model in the UniversalAI setup configuration, as long as we document why we're doing so and we agree on refactoring this part. Neither are a clean solution because
- `ExecutionWrapper` communicates with the components trhough the `IExecution` and `IComponent` interfaces, meaning it shouldn't care what each implementation does under the hood. By re-creating the usage handler we're breaking that principle, we only do so because one implementation alters the behaviour of one of the interface methods during the execution.
- Putting the method property in Setup isn't clean because the setup is meant to hold the connection details and Model has nothing to do with that. Moreover, Model is tied to the input property Task and now we're putting them in different parts of the spec.
- Neither solution addresses the issue where each job in a batch has different models or vendors to execute a task.

pkg/component/store/store.go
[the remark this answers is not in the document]
chuang8511: >   If you have several vendors, I guess you can call conn.WithInstillCredentials several times, which is a possible solution.

Yes, so the comment, "// Please apply more keys when we add more vendors", means we need to init more keys when we add more vendors.

>  In my comment I was wondering if it made sense to pass secrets as a whole in this case, or another strategy (I don't like passing unnecessary secrets to the component).

Ah, got you! I think it is a good preference! 👍 
I will keep them in my mind when we refactor those in the future!

chuang8511:
> ExecutionWrapper communicates with the components trhough the IExecution and IComponent interfaces, meaning it shouldn't care what each implementation does under the hood. By re-creating the usage handler we're breaking that principle, we only do so because one implementation alters the behaviour of one of the interface methods during the execution.
Putting the method property in Setup isn't clean because the setup is meant to hold the connection details and Model has nothing to do with that. Moreover, Model is tied to the input property Task and now we're putting them in different parts of the spec.
Neither solution addresses the issue where each job in a batch has different models or vendors to execute a task.

Got you! It is really clear! Thank you!

pkg/component/store/store.go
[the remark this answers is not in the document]
chuang8511: Recorded in ticket ins-5956.

pkg/component/ai/universalai/v0/config/tasks.json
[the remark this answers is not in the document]
chuang8511: move to 0.2.0

chuang8511:
About the to-be-refactored point, I noted them in to the secion of `Note about the future refactor` in ins-5956.

PR #681, opened 2024-09-30
chore(instillapp): init instill app

jvallesm:
This is a port of https://github.com/instill-ai/component/pull/366

Because

- we will have instill app

This commit

- init instill app


chuang8511:
I will move it to https://github.com/instill-ai/pipeline-backend/pull/670

PR #682, opened 2024-09-30
feat(text): add client sample code

jvallesm:
Because

This commit

### DIFF
--- a/pkg/component/example/text/main.go
+++ b/pkg/component/example/text/main.go
@@ -1,0 +1,60 @@
+package example
+
+import (
+	"flag"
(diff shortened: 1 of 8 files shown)


chuang8511:
I think the way I did before is not so good because it creases the cost of maintenance.
We should automate this part. 

Probably need to deal with this with the task that automates the sample recipe integration test.

chuang8511:
I think the way I did before is not so good because it creases the cost of maintenance.
We should automate this part.

Probably need to deal with this with the task that automates the sample recipe integration test.

PR #693, opened 2024-10-02
feat: add `range` support for iterator implementation

donch1989:
Because

- Currently, our iterator only implements the `foreach` function, meaning we iterate over the iterator `input`. This approach is insufficient for some use cases, where users may want to control the iteration's start and end points.

This commit

- Adds `range` support to the iterator implementation. It introduces the syntax `range: [start, stop [,step]]` to allow users to configure the iteration parameters. Additionally, users can specify the iteration index using `index: i`, with `i` being the default identifier.

### DIFF
--- a/pkg/datamodel/datamodel.go
+++ b/pkg/datamodel/datamodel.go
@@ -318,6 +318,8 @@ type Component struct {
 	Definition *Definition `json:"definition,omitempty" yaml:"-"`
 
 	// Fields for iterators
+	Range             []int                 `json:"range,omitempty" yaml:"range,omitempty"`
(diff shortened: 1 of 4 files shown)


donch1989:
Example recipe

```
variable:
  vars:
    instill-format: array:string

output:
  v:
    value: ${iterator-0.output.o1}

component:
  iterator-0:
    type: iterator
    range: [0,4,1]
    index: i
    component:
      base64-0:
        type: base64
        input:
          data: ${variable.vars[i]}
        task: TASK_ENCODE      
    output-elements:
      o1: ${base64-0.output.data}
```

PR #699, opened 2024-10-04
feat(integration): add identity to OAuth connections

jvallesm:
Because

- When a user authorizes an Instill component using OAuth, we want to
display the identity that they used.

This commit

- Adds identity information to OAuth connections.
- Makes OAuth method the only accepted one for connections if it's
  supported.

### DIFF
--- a/pkg/service/pipeline.go
+++ b/pkg/service/pipeline.go
@@ -1885,7 +1885,7 @@ func (s *service) ListPipelineRuns(ctx context.Context, req *pipelinepb.ListPipe
 
 	return &pipelinepb.ListPipelineRunsResponse{
 		PipelineRuns: pbPipelineRuns,
-		TotalSize:    totalCount,
+		TotalSize:    int32(totalCount),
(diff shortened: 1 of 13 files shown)


pkg/service/pipeline.go
```diff
@@ -1985,7 +1985,7 @@ func (s *service) ListComponentRuns(ctx context.Context, req *pipelinepb.ListCom
 
 	return &pipelinepb.ListComponentRunsResponse{
 		ComponentRuns: pbComponentRuns,
-		TotalSize:     totalCount,
+		TotalSize:     int32(totalCount),
```
jvallesm, line 1988: @donch1989 I didn't overthink the changes in this commit, I just wanted to fix the build after the breaking changes in https://github.com/instill-ai/protobufs/commit/e04c85940bfb1927202afe614cdda9623b5f5b44 and https://github.com/instill-ai/protobufs/commit/9f778f002b7289e0900aea75788e299be014ec7c.

PR #710, opened 2024-10-07
chore(component): update doc for instill app and universal ai components

chuang8511:
Because

- we have new components and want users to use them

This commit

- add the recipe sample for users


chuang8511:
The recipe should be like this.

image

GeorgeWilliamStrong (changes requested):
Please fix this typo that has come back

CleanShot 2024-10-15 at 12 46 04

chuang8511:
fixed by rebase

donch1989 (approved):

PR #726, opened 2024-10-10
perf(email): make `fetchEmails` async

donno2048:
solves instill-ai/instill-core#1111

### DIFF
--- a/pkg/component/application/email/v0/README.mdx
+++ b/pkg/component/application/email/v0/README.mdx
@@ -159,6 +159,7 @@ The search criteria for the emails
 | To | `to` | array | The email addresses of the recipient |
 </div>
 </details>
+
 #### Mailbox
 
 You have to confirm what exactly the mailbox name is.
--- a/pkg/component/application/email/v0/read_emails.go
+++ b/pkg/component/application/email/v0/read_emails.go
@@ -181,7 +181,7 @@ func fetchEmails(c *imapclient.Client, search Search) ([]Email, error) {
 		return []Email{}, nil
 	}
 
-	emails := make([]Email, cmp.Or(search.Limit, EmailReadingDefaultCapacity))
+	emails := make([]Email, cmp.Or(search.Limit, EmailReadingDefaultCapacity) + 1)
 	ch := make(chan Email)
 	errChan := make(chan error)
 	fetchOptions := &imap.FetchOptions{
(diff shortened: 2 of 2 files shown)


chuang8511:
@donno2048 
Thanks for your contribution.

## Summary
Please fix the "1." & "2." cases.


## Not as Expected Cases
The case with search case is not as expected.
1. The order is changed. We expected to fetch the email from latest email. But, it is from the earliest email.
Screenshot 2024-10-16 at 2 53 46 PM

2. Some content disappeared
Screenshot 2024-10-16 at 2 53 39 PM


## Good Cases
The case without search setting works fine, and the performance is good!

The end to end test with search limit: 100
originally
`"M":"TriggerPipelineWorkflow completed in","duration":"34.519138682s"`

After this PR merged
`"M":"TriggerPipelineWorkflow completed in","duration":"27.063491595s"`

image

donno2048:
Fixed "1." will work on "2." tomorrow

donno2048:
@chuang8511 please review

chuang8511:
@donno2048 
I guess the output data is not set. Could you confirm it again?
image

donno2048:
@chuang8511 Could you recheck it please?

chuang8511:
@donno2048 

The order seems still wrong.

image

Would you mind to do QA before requesting review?
I will check these [3 cases](https://github.com/instill-ai/pipeline-backend/pull/726#issuecomment-2416927049) as I mentioned before.

Thank you!

donno2048:
The problem is I don't have a PC and I can't really do that on mobile... Sorry...

chuang8511:
> The problem is I don't have a PC and I can't really do that on mobile... Sorry...

Sorry, just out of curiosity. You always develop in mobile?
Sound crazy cool. 😆 

OK, then let me QA for you. Please check your code again.

donno2048:
> > The problem is I don't have a PC and I can't really do that on mobile... Sorry...
> 
> Sorry, just out of curiosity. You always develop in mobile?


Yes 😅

chuang8511:
The code is not able to compile again.

Could you please find some ways to verify your code is ok?

Thank you. 🙏

donno2048:
Made a stupid mistake it'll now compile @chuang8511

chuang8511:
@donno2048 
I tested it again, but the code doesn't work. 😢 
- The order is still wrong
- When I set limit to 30, it runs forever

Please notify us to review when you find a way to test it.

The current test code is only unit testing for the  unexported functions in `email` pkg.
If you don't have PC, probably you can write some test code to exported functions to verify your modification.

donno2048:
I don't really have a way to do this right now, closing.


Thanks anyways

PR #727, opened 2024-10-10
feat(instillmodel): implement instill model embedding

chuang8511:
Because

- we need embedding from instill model

This commit

- add the standardised input & output for embedding
- add embedding to instill model
- support oneOf for the array items

### DIFF
--- a/pkg/component/ai/embedding.go
+++ b/pkg/component/ai/embedding.go
@@ -1,0 +1,39 @@
+package ai
+
+type EmbeddingInput struct {
+	Data      EmbeddingInputData `json:"data"`
+	Parameter EmbeddingParameter `json:"parameter"`
+}
+
+type EmbeddingInputData struct {
+	Model      string           `json:"model"`
+	Embeddings []InputEmbedding `json:"embeddings"`
+}
+
+type InputEmbedding struct {
+	Type        string `json:"type"`
+	Text        string `json:"text"`
+	ImageURL    string `json:"image-url"`
+	ImageBase64 string `json:"image-base64"`
+}
+
+type EmbeddingParameter struct {
+	Format     string `json:"format"`
+	Dimensions int    `json:"dimensions"`
+	InputType  string `json:"input-type"`
+	Truncate   string `json:"truncate"`
+}
+
+type EmbeddingOutput struct {
+	Data EmbeddingOutputData `json:"data"`
+}
+
+type EmbeddingOutputData struct {
+	Embeddings []OutputEmbedding `json:"embeddings"`
+}
+
+type OutputEmbedding struct {
+	Index   int   `json:"index"`
+	Vector  []any `json:"vector"`
+	Created int   `json:"created"`
+}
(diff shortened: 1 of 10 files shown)


chuang8511:
note, There is bug in model-backend. Before we merge it, we should wait for the bug fix and do end-to-end test again. Before that, we can review it first.

pkg/component/ai/embedding.go
```diff
@@ -0,0 +1,39 @@
+package ai
+
+type EmbeddingInput struct {
```
jvallesm, line 3: nit: please, document exported types. Although it doesn't produce a `golangci-lint` error / warning, many IDEs do show a warning here (and it hurts my eyes 🥲).

pkg/component/ai/instill/v0/embedding.go
```diff
@@ -0,0 +1,37 @@
...
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"google.golang.org/grpc/metadata"
+	"google.golang.org/protobuf/types/known/structpb"
+
+	modelPB "github.com/instill-ai/protogen-go/model/model/v1alpha"
```
jvallesm, line 11: nit: go packages aren't capitalised and aliases should follow that convention
```suggestion
	modelpb "github.com/instill-ai/protogen-go/model/model/v1alpha"
```

pkg/component/ai/instill/v0/main.go
```diff
@@ -101,12 +102,46 @@ func (e *execution) Execute(ctx context.Context, jobs []*base.Job) error {
 		defer gRPCCientConn.Close()
 	}
 
+	var result []*structpb.Struct
+
+	// We will refactor the component soon to align the data structure with Instill Model.
+	// For now, we move out `TASK_EMBEDDING` to a separate section.
```
jvallesm, line 108: 👍 Thanks for documenting this, it helps understand the code when you're visiting the package and you're not familiar with it

pkg/component/tools/compogen/pkg/gen/schema.go
```diff
@@ -15,6 +15,7 @@ type property struct {
 	Items struct {
 		Type       string              `json:"type"`
 		Properties map[string]property `json:"properties" validate:"omitempty,dive"`
+		OneOf []objectSchema `json:"oneOf" validate:"dive"`
```
jvallesm, line 18: nit: it looks like `gofmt` hasn't been executed (typically it does when you save a file) on this struct 🤔

jvallesm (approved):
Descriptions are a bit long but they're well-written and understandable. I added some non-blocking comments but the code looks good 👍

chuang8511:
> Descriptions are a bit long.

Yes, I intend to do that.
According to web operator experience, I think now we have the tendency to write the document in this pattern. I hope it aligns with what you think.

cc @GeorgeWilliamStrong

jvallesm (dismissed):
> Yes, I intend to do that.
According to web operator experience, I think now we have the tendency to write the document in this pattern. I hope it aligns with what you think.

Yes, no issues with it. I just noticed different from other components, where we tend to be more concise. But if it responds to a need we've observed, as long as descriptions are clear and well-written, I'm for it.

There is a conflict that should be resolved when generating the README doc. Pre-approving as the code is ready to be merged.

PR #742, opened 2024-10-15
fix(web): fix bug

chuang8511:
Because

- miss to update in go struct

This commit

- update go struct json

### DIFF
--- a/pkg/component/operator/web/v0/README.mdx
+++ b/pkg/component/operator/web/v0/README.mdx
@@ -32,7 +32,7 @@ This task involves systematically navigating through a website, starting from a 
 | Input | ID | Type | Description |
 | :--- | :--- | :--- | :--- |
 | Task ID (required) | `task` | string | `TASK_CRAWL_SITE` |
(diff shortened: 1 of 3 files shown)


chuang8511:
```
# VDP Version
version: v1beta

variable:
  input:
    instill-format: string

component:
  crawler:
    type: web
    input:
      url: ${variable.input}
      allowed-domains:
      max-k: 10
      timeout: 1000
      max-depth: 0
    condition:
    task: TASK_CRAWL_SITE

  scraper-iterator:
    type: iterator  
    input: ${crawler.output.pages}
    component:
      scraper:
        type: web
        input:
          url: ${scraper-iterator.element.link}
          include-html: false
          only-main-content: true
          remove-tags:
          only-include-tags:
          timeout: 0
        condition:
        task: TASK_SCRAPE_PAGE      
    output-elements:
      result: ${scraper.output}

output:
  result-pages:
    title: result pages
    value: ${scraper-iterator.output}
  result:
    title: result
    value: ${crawler.output.pages}

```

PR #748, opened 2024-10-16
feat(collection): add concat

chuang8511:
Because

- we need to concat 2 arrays

This commit

- add concat task in collection
- add recipe samples


pkg/component/generic/collection/v0/.compogen/bottom.mdx
```diff
@@ -0,0 +1,64 @@
+
```
jvallesm, line 1: As mentioned in another PR, instead of adding newlines we should put that in the `compogen` template. This is a workaround and contributors (myself included) don't need to know about this.

pkg/component/generic/collection/v0/.compogen/bottom.mdx
```diff
@@ -0,0 +1,64 @@
...
+
+## Example Recipes
+
+In this example, we are combining two arrays of strings using the `union` and `concat` components.
```
jvallesm, line 5: nit

```suggestion
### Combine 2 arrays using `TASK_UNION` and `TASK_CONCAT`

```

pkg/component/generic/collection/v0/.compogen/bottom.mdx
```diff
@@ -0,0 +1,64 @@
...
+  union-result:
+    title: Union Result
+    value: ${union.output.set}
+  concat-result:
+    title: Concat Result
+    value: ${concat.output.array}
+```
+
+
+In this example, we are creating an object with keys `type` and `text` from a string.
```
jvallesm, line 43: nit
```suggestion
### Use `TASK_ASSIGN` to create an object with `type` and `text` keys

```

pkg/component/generic/collection/v0/.compogen/bottom.mdx
```diff
@@ -0,0 +1,64 @@
+
+
+## Example Recipes
+
+In this example, we are combining two arrays of strings using the `union` and `concat` components.
+```yaml
+variable:
```
jvallesm, line 7: How about documenting and providing examples with a comment here? E.g.

```yaml
# This pipeline combines 2 arrays of strings
# Examples:
# ["foo", "bar"], ["zot", "bat"] -> ["foo", "bar", "zot", "bat"], []

File: pkg/component/generic/collection/v0/config/tasks.json [hunk 5]
```diff
@@ -316,5 +316,62 @@
       "title": "Output",
       "type": "object"
     }
+  },
+  "TASK_CONCAT": {
+    "instillShortDescription": "Concat the arrays. i.e. `[1, 2] + [3, 4] = [1, 2, 3, 4]`",
```
jvallesm, line 321:
```suggestion
    "instillShortDescription": "Concatenate the arrays. i.e. `[1, 2] + [3, 4] = [1, 2, 3, 4]`",
```

pkg/component/generic/collection/v0/config/tasks.json
```diff
@@ -316,5 +316,62 @@
...
+    "input": {
+      "description": "Input",
+      "instillEditOnNodeFields": [
+        "arrays"
+      ],
+      "instillUIOrder": 0,
+      "properties": {
+        "arrays": {
+          "description": "Specify the arrays you want to concat.",
```
jvallesm, line 330: nit: this should be a description, not an order
```suggestion
          "description": "The arrays to be concatenated.",
```

jvallesm: 🗣️ not crucial for this PR, just opening a discussion

We need to decide about the periods for the task tables. 2 options:
- Always finish with a period
- Don't finish with a period by default, only do so when we have > 1 sentence in the description (separated by a period).

cc @GeorgeWilliamStrong

pkg/component/generic/collection/v0/config/tasks.json
```diff
@@ -316,5 +316,62 @@
...
+    },
+    "output": {
+      "description": "Output",
+      "instillEditOnNodeFields": [
+        "array"
+      ],
+      "instillUIOrder": 0,
+      "properties": {
+        "array": {
+          "description": "The union set.",
```
jvallesm, line 361: This is a typo, right?
```suggestion
          "description": "The concatenated arrays.",
```

pkg/component/generic/collection/v0/main.go
```diff
@@ -82,6 +85,19 @@ func (c *component) CreateExecution(x base.ComponentExecution) (base.IExecution,
 	return e, nil
 }
 
+func (e *execution) concat(in *structpb.Struct) (*structpb.Struct, error) {
```
jvallesm, line 88: nit: unit-testing this component would be great 🙏

jvallesm (commented):
I added some suggestions for the documentation. Also some nitpicky comments, some things to consider (and perhaps address in a separate PR).

pkg/component/generic/collection/v0/.compogen/bottom.mdx
chuang8511: I remove `-` in compogen when there is extra content

pkg/component/generic/collection/v0/.compogen/bottom.mdx
chuang8511: I added these 2 in corresponding components.
```
# This component concatenates two arrays of strings.
# Examples:
# ["foo", "bar"], ["zot", "bat"] -> ["foo", "bar", "zot", "bat"]
```


```
# This component combines two arrays of strings in a union, ensuring that no duplicates are included.
# Examples:
# ["foo", "bar"], ["foo", "bat"] -> ["foo", "bar", "bat"]
```

pkg/component/generic/collection/v0/config/tasks.json
chuang8511: > Always finish with a period

Now, I follow this, which @GeorgeWilliamStrong mentioned it somewhere.

chuang8511:
@jvallesm 
Thanks for reviewing. I fixed all and made a few adjustments.
Please take a look. 

And, I am curious about if there is a more elegant way to init the test data in your experience.

I felt [it](https://github.com/instill-ai/pipeline-backend/pull/748/files#diff-b6973a993567c10b1d03b3bc80e255bd6a816d14197950ed6fd930788a205955R18) is not so easy to create the test data.

donch1989 (approved):
It’s odd that `compongen` generates empty lines at the end of MDX files, but the pre-commit action removes them. I’ve adjusted the pre-commit hook to fix this behavior so the end-of-line characters in MDX files are no longer automatically removed. We can decide whether we should keep empty lines at the end of MDX files later.

jvallesm:
> It’s odd that `compongen` generates empty lines at the end of MDX files, but the pre-commit action removes them. I’ve adjusted the pre-commit hook to fix this behavior so the end-of-line characters in MDX files are no longer automatically removed. We can decide whether we should keep empty lines at the end of MDX files later.

It's a compromise between the readability of the template and the auto-generated file. Since the template ends with [an optional section](https://github.com/instill-ai/pipeline-backend/blob/main/pkg/component/tools/compogen/pkg/gen/resources/templates/readme.mdx.tmpl#L188), this will produce a blank line by default (empty section). My opinion is that, while we should aim to make README.mdx as pretty as possible, the shape of autogenerated files isn't important as long as their rendered version (or e.g. the API of the package for auto-generated code) is neat. When in conflict we should prioritise rendered document > template > raw MDX.

jvallesm:
> And, I am curious about if there is a more elegant way to init the test data in your experience.
> 
> I felt [it](https://github.com/instill-ai/pipeline-backend/pull/748/files#diff-b6973a993567c10b1d03b3bc80e255bd6a816d14197950ed6fd930788a205955R18) is not so easy to create the test data.

I can think of 2 simpler ways to define the test data:
- As Go types (`[]string{"c", "d"}`, `map[string]any{"array": want}`). [`NewValue`](https://pkg.go.dev/google.golang.org/protobuf/types/known/structpb#NewValue) takes any type and will do the conversion for you.
- You can also define the test data in JSON and use `protojson` to unmarshal it.

pkg/component/generic/collection/v0/.compogen/bottom.mdx
jvallesm: I meant documented the input/output of the **pipelines**, not the individual components. But it's fine.

PR #750, opened 2024-10-16
ci(Dockerfile,Makefile): refine targets and fix warnings

pinglin:
Because

- there are confusing and unuseful Makefile targets
- the latest Docker version warns ENV syntax and default ARG value

This commit

- fixes the issues
- updates CODEOWNERS


Dockerfile
```diff
@@ -1,4 +1,5 @@
-FROM --platform=$TARGETPLATFORM golang:1.22.5-alpine3.19 AS build
```
chuang8511, line 1: I don't know what `$TARGETPLATFORM` is using for.

chuang8511 (commented):
Only 1 thing to confirm.
If it is ok, others LGTM.

PR #751, opened 2024-10-16
chore(component): update doc

chuang8511:
Because

- the doc is unclear

This commit

- update doc

### DIFF
--- a/pkg/component/application/email/v0/.compogen/extra-intro.mdx
+++ b/pkg/component/application/email/v0/.compogen/extra-intro.mdx
@@ -1,2 +1,4 @@
+
+
 You can connect to different email servers through the Email component.
 Emails are fetched and sent using the IMAP and SMTP protocols, respectively. You can set the server address and port for each protocol in the component configuration.
--- a/pkg/component/application/email/v0/.compogen/extra-intro.mdx
+++ b/pkg/component/application/email/v0/.compogen/extra-intro.mdx
@@ -1,4 +1,2 @@
-
-
 You can connect to different email servers through the Email component.
 Emails are fetched and sent using the IMAP and SMTP protocols, respectively. You can set the server address and port for each protocol in the component configuration.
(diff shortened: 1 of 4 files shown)


pkg/component/application/email/v0/.compogen/extra-intro.mdx
```diff
@@ -1,2 +1,4 @@
+
```
jvallesm, line 1: This will need to be added to every extra content appended to the intro. Why not adding the newlines in the template instead?

pkg/component/data/instillartifact/v0/config/tasks.json
```diff
@@ -184,7 +184,7 @@
                   "$ref": "#/$defs/file"
                 },
                 "file-name": {
-                  "description": "The name of the file, please remember to add the file extension in the end of file name. e.g. 'example.pdf'",
+                  "description": "The name of the file, please remember to add the file extension in the end of file name. e.g. 'example.pdf'. The length limitation of file name is 100 characters.",
```
jvallesm, line 187: nit: 
- There are slight differences in style for the different field descriptions. In this case, this is the only description that stars by "The ..." which produces a small style inconsistency. 
- Comma isn't the best separator between the definition of the field the part about the extension.

```suggestion
                  "description": "Name of the file, including the extension (e.g. `example.pdf`). The length of this field is limited to 100 characters.",
```

pkg/component/application/email/v0/.compogen/extra-intro.mdx
chuang8511: I just thought that we tend to clean the blank line by [this way](https://github.com/instill-ai/pipeline-backend/blob/main/pkg/component/tools/compogen/pkg/gen/resources/templates/readme.mdx.tmpl#L13).
So, I did not want to touch template again.

So, do you mean I just take out `-` in the template?

chuang8511: modify it in another branch.

chuang8511:
To avoid conflict with other branch, I did not execute `make gen-component-doc`.

jvallesm (dismissed):
Now things are merged, I think you should be able to re-generate the docs.

PR #753, opened 2024-10-17
feat(web): refactor web operator

chuang8511:
Because

- we want to increase max k
- we want to make scrape method explicit

This commit

- refactor crawler
- add scrape method for scraper

### DIFF
--- a/pkg/component/operator/web/v0/config/tasks.json
+++ b/pkg/component/operator/web/v0/config/tasks.json
@@ -62,8 +62,8 @@
           "type": "array"
         },
         "max-k": {
-          "default": 1,
-          "description": "Max-K sets a limit on the number of pages to fetch. If Max-K is large and all pages cannot be fetched within the time limit, the fetching process will continue for up to 120 seconds. If Max-K is a positive number, the fetch will return up to that many pages, but no more.",
+          "default": 10,
+          "description": "Max-K sets a limit on the number of pages to fetch. If Max-K is set to 0, all available pages will be fetched within the time limit of 120 seconds. If Max-K is a positive number, the fetch will return up to that many pages, but no more.",
           "instillAcceptFormats": [
             "integer"
           ],
@@ -72,7 +72,7 @@
             "value",
             "reference"
           ],
-          "minimum": 1,
+          "minimum": 0,
           "title": "Max Number of Pages",
           "type": "integer"
         },
--- a/pkg/component/operator/web/v0/config/tasks.json
+++ b/pkg/component/operator/web/v0/config/tasks.json
@@ -62,8 +62,8 @@
           "type": "array"
         },
         "max-k": {
-          "default": 10,
-          "description": "Max-K specifies the maximum number of pages to return. If max-k is set to 0, all available pages will be returned, up to a maximum of 100. If max-k is set to a positive number, the result will include up to max-k pages, but no more than that.",
+          "default": 1,
+          "description": "Max-K sets a limit on the number of pages to fetch. If Max-K is large and all pages cannot be fetched within the time limit, the fetching process will continue for up to 120 seconds. If Max-K is a positive number, the fetch will return up to that many pages, but no more.",
           "instillAcceptFormats": [
             "integer"
           ],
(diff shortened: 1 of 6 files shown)


chuang8511:
QA cases
image

pkg/component/operator/web/v0/crawl_website.go
```diff
@@ -70,45 +76,67 @@ func (e *execution) CrawlWebsite(input *structpb.Struct) (*structpb.Struct, erro
...
 
+	// We will have the component timeout feature in the future.
+	// Before that, we initialize the context here.
+	ctx, cancel := context.WithTimeout(context.Background(), 120*time.Second)
+	defer cancel()
+
 	// On every a element which has href attribute call callback
 	// Wont be called if error occurs
 	c.OnHTML("a[href]", func(e *colly.HTMLElement) {
+		mu.Lock()
```
donch1989, line 87:
```suggestion
		mu.Lock()
                defer mu.Unlock()
```
The mutex need to be unlocked no matter what. We can use `defer` to guard this. I assume the reason you didn’t use `defer` is that you didn’t want to scope `_ = e.Request.Visit(link)` within the mutex. Perhaps we can rearrange the code to avoid this issue while still utilizing `defer`.

pkg/component/operator/web/v0/crawl_website.go
```diff
@@ -128,20 +156,36 @@ func (e *execution) CrawlWebsite(input *structpb.Struct) (*structpb.Struct, erro
 		title := util.ScrapeWebpageTitle(doc)
 		page.Title = title
 
-		defer mu.Unlock()
```
donch1989, line 131: Any reason why we don't use `defer` here?

pkg/component/operator/web/v0/crawl_website.go
```diff
@@ -128,20 +156,36 @@ func (e *execution) CrawlWebsite(input *structpb.Struct) (*structpb.Struct, erro
...
 
 	// Start scraping
 	if !strings.HasPrefix(inputStruct.URL, "http://") && !strings.HasPrefix(inputStruct.URL, "https://") {
 		inputStruct.URL = "https://" + inputStruct.URL
 	}
-	_ = c.Visit(inputStruct.URL)
-	c.Wait()
+
+	go func() {
+		_ = c.Visit(inputStruct.URL)
```
donch1989, line 184: Since it's in a goroutine now. Do we still need to `c.Wait()` here?

donch1989 (changes requested):

pkg/component/operator/web/v0/config/tasks.json
```diff
@@ -63,7 +63,7 @@
         },
         "max-k": {
           "default": 10,
-          "description": "Max-K specifies the maximum number of pages to return. If max-k is set to 0, all available pages will be returned, up to a maximum of 100. If max-k is set to a positive number, the result will include up to max-k pages, but no more than that.",
+          "description": "Max-K sets a limit on the number of pages to fetch. If Max-K is set to 0, all available pages will be fetched within the time limit of 120 seconds. If Max-K is a positive number, the fetch will return up to that many pages, but no more.",
```
jvallesm, line 66: I think we should document the 120s limitation on the task description, not the max-k field. Though I value having a note here saying that when the task timeout is reached the available results are returned.

pkg/component/operator/web/v0/crawl_website.go
```diff
@@ -38,9 +41,12 @@ type CrawlWebsiteInput struct {
...
 
-func (i *CrawlWebsiteInput) Preset() {
-	if i.MaxK < 0 {
-		i.MaxK = 0
+func (i *CrawlWebsiteInput) preset() {
+	if i.MaxK <= 0 {
+		// When the users set to 0, it means infinite.
+		// However, there is performance issue when we set it to infinite.
+		// So, we set the default value to solve performance issue easily.
+		i.MaxK = 8000
```
jvallesm, line 49: This limitation should be documented in the public readme.

donch1989:
I'll merge this first. @chuang8511 please check the comments and send another PR later.

pkg/component/operator/web/v0/crawl_website.go
chuang8511: The colly is hard to understand.
So, I just tried lots of workaround to confirm if it works.

I will clean the code and make sure the performance is same.

pkg/component/operator/web/v0/crawl_website.go
chuang8511: Yes, we need it. `c.Wait()` is for async crawling in colly package.

pkg/component/operator/web/v0/crawl_website.go
chuang8511: Do you mean to make it in `README.mdx`?

This setting is just to fix the performance issue. So, the users don't have to know this information. Mainly, it is for developer. I will add clearer comments here.

chuang8511: I updated [here](https://github.com/instill-ai/pipeline-backend/pull/772/commits/c120df686c18a8618f0d565be751f0c2c7e89729).

PR #762, opened 2024-10-20
feat(component,audio): add TASK_DETECT_ACTIVITY and TASK_SEGMENT

pinglin:
Because

- Voice activity detection (VAD) is an essential feature for audio and video segmentation, enabling further text transcription for RAG embedding and LLM understanding.

This commit

- adds `TASK_DETECT_ACTIVITY`
- adds `TASK_SEGMENT`
- refactors the previous Audio operator
  - removing `TASK_CHUNK_AUDIOS`
   - refactoring `TASK_SLICE_AUDIO` to `TASK_SEGMENT`
- makes the production image use base image `debian:bullseye-slim` because `onnxruntime` doesn't support Alpine apk and we probably don't wanna build it from scratch


pkg/component/operator/audio/v0/task_detect_voice_activity.go
```diff
@@ -0,0 +1,130 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("detect voice activity: %w", err)
+	}
+
+	segments, err := sd.Detect(audioBuf.AsFloat32Buffer().Data)
+	if err != nil {
+		return nil, fmt.Errorf("detect voice activity: %w", err)
+	}
+
+	if err := sd.Destroy(); err != nil {
```
jvallesm, line 91: ❓ Can we use `defer` for this?

pkg/component/operator/audio/v0/audio.go
```diff
@@ -0,0 +1,13 @@
+package audio
+
+type Audio string
```
jvallesm, line 3: I suspect this is an exported type because it conflicts with the `audio` package, correct me if I'm wrong.
- Can we use another alias so the type is unexported?
- Do we need the alias at all? The type doesn't implement any extra methods over `string`.

pkg/component/operator/audio/v0/task_detect_voice_activity_test.go
jvallesm, line 1: 👍 Nice coverage, thanks for adding it.

nit: I prefer to test the public API of the package if possible, testing [the output of the `Execute` method](https://github.com/instill-ai/pipeline-backend/blob/main/pkg/component/operator/json/v0/component_test.go#L211) rather than the private executors. But this is a very slight change as `Execute` thinly wraps the methods you're testing, so your approach is valid to me, too.

.github/workflows/coverage.yml
jvallesm, line 1: The dev environment will also need `onnxruntime`, with the current changes `make test` won't work. I'd:
- Document this dependency in the contribution guidelines / in the `audio` package tests
- Make the `audio` tests optional by adding a build flag like in `operator/document`. Then a flag can be passed to the `make test` and `make coverage` commands like `OCR=true` to always run it on the CI.

jvallesm (commented):
Production code looks good, I added some comments to improve the documentation and tests.

pkg/component/operator/audio/v0/task_extract_audio_segments_test.go
```diff
@@ -0,0 +1,60 @@
...
+		audioFile     string
+		segmentsFile  string
+		expectedCount int
+	}{
+		audioFile:     "testdata/voice.wav",
+		segmentsFile:  "testdata/voice-activity-segments.json",
+		expectedCount: 15,
+	}
+
+	// Load audio file
```
chuang8511, line 28: I think good comments are for
1. Exported document
2. Reason why the code is like this
3. If it is hard to understand, add what the code does

The code explains itself well, so I'd remove the comments here.

chuang8511 (commented):
Leave some questions and comments.

Dockerfile
```diff
@@ -1,59 +1,60 @@
 ARG GOLANG_VERSION=1.22.5
-FROM golang:${GOLANG_VERSION}-alpine3.19 AS build
-
-RUN apk add --no-cache build-base leptonica-dev tesseract-ocr-dev musl-dev
+FROM golang:${GOLANG_VERSION}-bullseye AS build
```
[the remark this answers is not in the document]
chuang8511: I saw the message in Slack.

pkg/component/operator/audio/v0/task_detect_voice_activity.go
pinglin: We actually want to destroy the model instance right after the inference has been performed.

pkg/component/operator/audio/v0/audio.go
pinglin: Will rename it to `base64Audio` to align with the Image operator too. These are temporary data structure anyhow before we have the binary data exchange implementation.

pkg/component/operator/audio/v0/task_detect_voice_activity_test.go
pinglin: Will refactor the testing to align with the style.

Dockerfile
```diff
@@ -1,59 +1,60 @@
...
-# Download tesseract data
-RUN curl -L https://github.com/tesseract-ocr/tessdata_best/raw/main/eng.traineddata \
-    -o /usr/share/tessdata/eng.traineddata
-
-ARG TARGETARCH
-ARG BUILDARCH
-RUN apk add unrtf --repository=http://dl-cdn.alpinelinux.org/alpine/edge/community
+# Install Python, create virtual environment, and install pdfplumber
+RUN apt update && \
+    apt install -y curl python3 python3-venv poppler-utils wv unrtf tidy tesseract-ocr libtesseract-dev libreoffice ffmpeg chromium qpdf && \
```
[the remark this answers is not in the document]
pinglin: The unit test and integration test are passed. Haven't manually tested them yet but I suggest we add missing unit and integration test for those by the chance.

pkg/component/operator/audio/v0/config/tasks.json
```diff
@@ -1,18 +1,41 @@
...
+        "start-time",
+        "end-time"
       ],
+      "title": "Segment",
+      "type": "object",
+      "description": "A segment of audio data"
+    }
+  },
+  "TASK_DETECT_VOICE_ACTIVITY": {
+    "instillShortDescription": "Detect voice activity in audio data. The audio data will be internally converted to sample rate 48kHz and a mono channel, if the original format is not 48kHz and mono.",
```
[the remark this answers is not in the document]
pinglin: Yes, will make it more accessible to common audience.

pkg/component/operator/audio/v0/config/tasks.json
```diff
@@ -1,18 +1,41 @@
...
+          "title": "End Time",
+          "type": "number",
+          "description": "The end time of the segment in seconds",
+          "instillFormat": "number",
+          "instillUIOrder": 1
+        }
+      },
+      "required": [
+        "start-time",
+        "end-time"
```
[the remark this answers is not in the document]
pinglin: `segment` is a general data structure, in which both `start-time` and `end-time` are required to define a time segment. The corner case of unbound end-time should be handled by the consumer of this data structure.

.github/workflows/coverage.yml
pinglin: The dev container has the `onnxruntime` set up actually. `make test` should work in the `make dev` container. Can you try again?

Will add the flag to make the test depending on `onnxruntime` optional.

jvallesm: Yeah, I meant more in the local dev environment. Being able to run `go test` while you develop (e.g. from the IDE directly) is a plus, so I'd make every test that has dependencies optional. 

I don't have a strong opinion about `make test`, that can be meant to be run in the container, at the moment OCR tests are disabled by default in that make command.

pinglin: Setting up TesseractOCR and ONNX Runtime at local dev env can be quite a burden indeed. Let's just ensure `make test OCR=true ONNX=true` working in the dev container for now.

pkg/component/operator/audio/v0/config/tasks.json
```diff
@@ -22,65 +45,68 @@
...
-        "audios"
+        "segments"
       ],
       "title": "Output",
       "type": "object"
     }
   },
-  "TASK_SLICE_AUDIO": {
-    "instillShortDescription": "Specify a time range to slice an audio file",
+  "TASK_EXTRACT_AUDIO_SEGMENTS": {
```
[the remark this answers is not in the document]
pinglin: As this PR will completely change the input API contract and refactor directly in `v0` folder, there is no trivial way to migrate the previous input code. Given the Audio operator is alpha and usage is still low, I suggest we do the breaking change and I will work on refactoring the Video operator for the same too.

pkg/component/operator/audio/v0/task_detect_activity_test.go
```diff
@@ -0,0 +1,150 @@
+//go:build onnx
+// +build onnx
+
+// This test file requires ONNX Runtime to be installed. Follow these steps to set it up:
```
jvallesm, line 4: ❤️ Thanks for adding the tags and documentation, they add a lot of clarity

pkg/component/operator/audio/v0/task_detect_activity.go
```diff
@@ -0,0 +1,125 @@
...
+	"context"
+	"encoding/base64"
+	"fmt"
+	"os"
+	"path/filepath"
+
+	"github.com/go-audio/audio"
+	"github.com/go-audio/wav"
+	"github.com/mjibson/go-dsp/fft"
+	"github.com/streamer45/silero-vad-go/speech"
```
jvallesm, line 14: This also depends on `ONNX` so building the code on the local dev env fails, which might create confusion. Bulding the code as I write it is a usual step to detect errors early.

```
go build ./cmd/main
# github.com/streamer45/silero-vad-go/speech
In file included from ../../../../pkg/mod/github.com/streamer45/silero-vad-go@v0.2.1/speech/detector.go:5:
./ort_bridge.h:1:10: fatal error: 'onnxruntime_c_api.h' file not found
    1 | #include "onnxruntime_c_api.h"
      |          ^~~~~~~~~~~~~~~~~~~~~
1 error generated.
```

pinglin: Right. The problematic `include` is all the way from the dependent package. Will add the build tag correspondingly.

pkg/component/operator/audio/v0/README.mdx
```diff
@@ -5,10 +5,10 @@ draft: false
 description: "Learn about how to set up a VDP Audio component https://github.com/instill-ai/instill-core"
 ---
 
-The Audio component is an operator component that allows users to extract and manipulate audio from different sources.
+The Audio component is an operator component that allows users to operate audio data..
```
GeorgeWilliamStrong, line 8:
```suggestion
The Audio component is an operator component that allows users to operate audio data.
```

GeorgeWilliamStrong (commented):
Left one minor suggestion regarding docs. Otherwise LGTM

pkg/component/operator/audio/v0/README.mdx
pinglin: The double periods have to do with the new description rule (i.e., always use period) resulting the doc generated repeating the period due to the template hasn't been updated (another PR will do this). But yes, this reminds me that we can actually just manually change this actually.

pinglin:
Hey @jvallesm, I spent some time simplifying `make test` by removing unnecessary complex logic from the `make test-ocr` target. Please give it a try and see if it works for you.

jvallesm (dismissed):
> > Hey @jvallesm, I spent some time simplifying `make test` by removing unnecessary complex logic from the `make test-ocr` target. Please give it a try and see if it works for you.

Yes, thanks for handling that. I can now build the code as I edit it and run `make test` locally 👍

PR #765, opened 2024-10-21
feat(component, video): implemented fractional fps value - fps as number, and added some debugging flags for ffmpeg

Reckadon:
Because

- The Video Component minimum FPS is 1 (integer), For a use case like presentation in the video, we do not want to extract 1 frame every second but 1 frame every 10 seconds to save resources and reduce the number of images needed to feed into the next component. 
- Issue instill-ai/instill-core/issues/1115.

This commit

- Change FPS to accept real number inputs as well i.e. 60, 30, 1, 0.5, 0.2, 0.1 in both the tasks `TASK_SUBSAMPLE_VIDEO` and `TASK_SUBSAMPLE_VIDEO_FRAMES`.
- Also implements a string parsing function which can handle fractional inputs like 1/2 and 1/5 but is commented out - because the input `type` can only be defined as one single type (`number` or `string` and not both). The function can be used if this issue is resolved in the future.

### DIFF
--- a/pkg/component/operator/video/v0/video_operation.go
+++ b/pkg/component/operator/video/v0/video_operation.go
@@ -149,16 +149,12 @@ func subsampleVideoFrames(input *structpb.Struct) (*structpb.Struct, error) {
 	// with frame number rather than uuid as suffix.
 	outputPattern := random + "_frame_%08d.jpeg"
 
-	kwArgs, errArgs := getFramesKwArgs(inputStruct)
-	if errArgs != nil {
-		return nil, fmt.Errorf("unable to convert fps to float number: %s", errArgs)
-	}
+	kwArgs := getFramesKwArgs(inputStruct)
 
 	err = ffmpeg.Input(tempInputFileName).
 		Output(outputPattern,
 			kwArgs,
 		).
-		//.GlobalArgs("-report").OverWriteOutput().
 		Run()
 
 	if err != nil {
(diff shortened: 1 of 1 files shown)


Reckadon:
Any updates or comments regarding the PR? @chuang8511 @jvallesm @donch1989

pkg/component/operator/video/v0/video_operation.go
```diff
@@ -186,19 +192,47 @@ func subsampleVideoFrames(input *structpb.Struct) (*structpb.Struct, error) {
...
+	return kwArgs, nil
 }
 
 func removeFiles(files []string) {
 	for _, file := range files {
 		os.Remove(file)
 	}
 }
+
+// func getFormattedFPS(fpsLiteral string) (float32, error) {
```
chuang8511, line 218: Please remove unnecessary code.

pkg/component/operator/video/v0/video_operation.go
```diff
@@ -186,19 +192,47 @@ func subsampleVideoFrames(input *structpb.Struct) (*structpb.Struct, error) {
 	return base.ConvertToStructpb(output)
 }
 
-func getFramesKwArgs(inputStruct SubsampleVideoFramesInput) ffmpeg.KwArgs {
-	kwArgs := ffmpeg.KwArgs{"vf": "fps=" + fmt.Sprintf("%d", inputStruct.Fps)}
+func getFramesKwArgs(inputStruct SubsampleVideoFramesInput) (ffmpeg.KwArgs, error) {
```
chuang8511, line 195: Why do we need to return error here?
It seems there is no case to return error

chuang8511 (commented):
Left a few comments.

And, I want to confirm - what you've done is only change fps types. right?

ffmpeg itself can receive variable with type float.
Is it correct?

chuang8511:
QAed

image


Tested Recipe
```
# VDP Version
version: v1beta

variable:
  video:
    title: Video
    instill-format: video
  fps:
    title: FPS
    instill-format: number

component:
  video-0:
    type: video
    input:
      video: ${variable.video}
      fps: ${variable.fps}
      start-time: 0:0:0
      duration: 0:0:12
    condition:
    task: TASK_SUBSAMPLE_VIDEO_FRAMES  

output:
  frames:
    title: Frames
    value: ${video-0.output.frames}

```

Reckadon:
> Left a few comments.
> 
> And, I want to confirm - what you've done is only change fps types. right?
> 
> ffmpeg itself can receive variable with type float. Is it correct?

Yes, due to the `string/number` issue, that was the only thing i did.

pkg/component/operator/video/v0/video_operation.go
Reckadon: left that in case a string processing function is implemented. will remove

Reckadon:
@chuang8511 i removed the unnecessary code.

chuang8511 (approved):
LGTM

PR #770, opened 2024-10-23
feat(run): add requester id in list pipeline run response

joremysh:
Because

- console runs page needs to display requester

This commit

- expose requester id in pipeline runs response

image

### DIFF
--- a/pkg/service/pipelinerun.go
+++ b/pkg/service/pipelinerun.go
@@ -119,18 +119,19 @@ func (s *service) ListPipelineRuns(ctx context.Context, req *pb.ListPipelineRuns
 		metadataMap[content.Name] = content.Content
 	}
 
-	requesterIDMap := make(map[string]struct{})
+	userUIDMap := make(map[string]struct{})
 	for _, pipelineRun := range pipelineRuns {
-		requesterIDMap[pipelineRun.TriggeredBy] = struct{}{}
+		userUIDMap[pipelineRun.TriggeredBy] = struct{}{}
+		userUIDMap[pipelineRun.Namespace] = struct{}{}
 	}
 
-	runnerMap := make(map[string]*string)
-	for requesterID := range requesterIDMap {
+	userIDMap := make(map[string]*string)
+	for requesterID := range userUIDMap {
 		runner, err := s.mgmtPrivateServiceClient.CheckNamespaceByUIDAdmin(ctx, &mgmtpb.CheckNamespaceByUIDAdminRequest{Uid: requesterID})
 		if err != nil {
 			return nil, err
 		}
-		runnerMap[requesterID] = &runner.Id
+		userIDMap[requesterID] = &runner.Id
 	}
 
 	// Convert datamodel.PipelineRun to pb.PipelineRun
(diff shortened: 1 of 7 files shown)


pkg/service/pipelinerun.go
```diff
@@ -140,7 +141,10 @@ func (s *service) ListPipelineRuns(ctx context.Context, req *pb.ListPipelineRuns
 		if err != nil {
 			return nil, fmt.Errorf("failed to convert pipeline run: %w", err)
 		}
-		pbRun.RunnerId = runnerMap[run.TriggeredBy]
+		pbRun.RunnerId = userIDMap[run.TriggeredBy]
+		if requesterID, ok := userIDMap[run.Namespace]; ok && requesterID != nil {
```
jvallesm, line 145: nit: Ideally I'd use the same strategy for both fields (empty string instead of `null` if the value doesn't exist). If you must use both approaches, I'd favour handling strings instead of pointers. Even if the code is simple, with the former you never risk accessing a `null` pointer.

pkg/service/pipelinerun.go
```diff
@@ -346,11 +359,15 @@ func (s *service) ListPipelineRunsByRequester(ctx context.Context, req *pb.ListP
 		if err != nil {
 			return nil, fmt.Errorf("converting pipeline run: %w", err)
 		}
-		pbRun.RunnerId = runnerMap[run.TriggeredBy]
+		pbRun.RunnerId = userIDMap[run.TriggeredBy]
```
jvallesm, line 362: nit: it seems we're maintaining the same code in 2 places. We might want to extract the common code to a function.

jvallesm (approved):
No blocking comments, I added some notes on improving clarity in the code.

pkg/service/pipelinerun.go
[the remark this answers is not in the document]
joremysh: The data table columns and data model fields should be renamed to make it more straightforward and align with those in model-backend. And I'm also considering to modify the column type from string to UUID. I would like to fix these after the renaming.

We can keep those fields as they are right now.

joremysh: That is reasonable but I don't think it should be included in this ticket. We could update the RunnerId to string instead of string pointer later.

pkg/service/pipelinerun.go
joremysh: I've thought of this but it will include an API call to mgmt private service. Did not come up with a good idea on this so I did not extract them yet.
Also I'm also considering adding a memcache package to cache the user IDs here.

pkg/service/pipelinerun.go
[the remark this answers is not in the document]
jvallesm: > I'm also considering to modify the column type from string to UUID

I noticed this, too, and it would definitely help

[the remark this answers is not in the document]
joremysh: For the data table fields, I am planning to rename them as follows:
`namespace` --> `requester_uid`
`triggered_by` --> `runner_uid`

PR #774, opened 2024-10-23
feat: revamp Instill Format

donch1989:
Because
 - Instill Format needs a more flexible, user-friendly way to handle diverse data types and conversions.
 - Binary data support and format standardization improve functionality and allow smoother developer and user experiences.

This commit
 - Introduces support for unstructured data formats, with auto-conversion for image, audio, and video codec.
 - Enables data attributes access in file-type data.
 - Adds support for default values in the variable section, allowing fields to be optionally omitted in requests if a default is set.
 - Adds URL-based file upload support with a flexible workflow for base64 and URL uploads.
 - Separates data structs and format interfaces, ensuring Instill Format effectively differentiates data content from usage intent.
 - Updates components (openai, anthropic, mistral, text, document, image, collection) to align with the new interface structures.


pkg/component/operator/image/v0/helper_test.go
```diff
@@ -0,0 +1,93 @@
...
+	imgData, err := data.NewImageFromBytes(buf.Bytes(), "image/png", fmt.Sprintf("image-%s.png", uuid.New().String()))
+	if err != nil {
+		c.Assert(err, qt.IsNil)
+	}
+	return imgData
+}
+
+func compareTestImage(c *qt.C, img format.Image, name string) {
+
+	fileName := fmt.Sprintf("testdata/test_output_%s_%s_%s.jpeg", name, strings.ToLower(strings.Split(c.Name(), "/")[0]), strings.ToLower(strings.Split(c.Name(), "/")[1]))
```
pinglin, line 47:
```suggestion
	filename := fmt.Sprintf("testdata/test_output_%s_%s_%s.jpeg", name, strings.ToLower(strings.Split(c.Name(), "/")[0]), strings.ToLower(strings.Split(c.Name(), "/")[1]))
```

.github/workflows/coverage.yml
```diff
@@ -42,7 +40,7 @@ jobs:
 
       - name: Generate coverage report
         run: |
-          go mod tidy
+          make build-dev
```
pinglin, line 43:
```suggestion
```

pkg/data/struct.go
```diff
@@ -0,0 +1,391 @@
...
+	if field.Type().Implements(reflect.TypeOf((*format.Value)(nil)).Elem()) {
+		field.Set(reflect.ValueOf(v))
+		return nil
+	}
+	return fmt.Errorf("cannot unmarshal %T into %v", v, field.Type())
+}
+
+// getFieldName returns the field name from the struct tag or the field name itself.
+func getFieldName(field reflect.StructField) string {
+	fieldName := field.Tag.Get("key")
```
pinglin, line 278:
```suggestion
	fieldName := field.Tag.Get("json")
```

pkg/data/document.go
```diff
@@ -4,91 +4,121 @@ import (
...
+			isSupported = true
+			break
+		}
+		if f.contentType == supportedType {
+			isSupported = true
+			break
+		}
+	}
+	fmt.Println()
+	fmt.Println(isSupported)
```
pinglin, line 72:
```suggestion
```

pkg/data/struct.go
donch1989: We are not using json, since it's no longer a json.
Instead, you need to use the `key` tag in your struct.

For example, 
```
type detectActivityInput struct {
	Audio              format.Audio `key:"audio"`
	MinSilenceDuration int          `key:"min-silence-duration"`
	SpeechPad          int          `key:"speech-pad"`
}

```

pinglin: That makes much more sense then. Great!

pkg/data/format/format.go
```diff
@@ -0,0 +1,82 @@
...
+
+type File interface {
+	Value
+
+	Binary() (ba ByteArray, err error)
+	DataURI() (url String, err error)
+	Base64() (url String, err error)
+	FileSize() (size Number)
+	ContentType() (t String)
+	FileName() (t String)
```
pinglin, line 47:
```suggestion
	Filename() (t String)
```

donch1989: I’m considering this issue as well. I think both `Filename` and `File Name` are valid terms, with `Filename` being more commonly used in programming languages.

However, regarding the data attributes we’ve provided, we have many attributes prefixed with `File`, such as `File Size`, `File Name`, and `File Format`. Should we standardize their naming for consistency? This would result in using FileName or file-name, FileSize or file-size, and FileFormat or file-format, but not `Filename` and `filename`.

What do you think?

PR #779, opened 2024-10-25
feat(minio): import updated minio package and add tag on file upload

joremysh:
Because

- run logging metadata retention

This commit

- import updated minio package and add tag on file upload


config/config.go
```diff
@@ -248,3 +239,14 @@ func ParseConfigFlag() string {
 
 	return *configPath
 }
+
+const (
+	FreePlanExpiryTag = "free-plan-expiry"
```
jvallesm, line 244: ❓ Can we use another tag (e.g. `default-expiry`) or is the logic tied to this constant? I think it'd be better if we could  avoid using a cloud-only term like "plan"

jvallesm (commented):
Looks good, just a small comment about terminology

jvallesm:
🙏 Can you remove the commit list from the PR description. It'll end up in the merge commit message. Thanks!

> 
    chore(minio): replace minio package with common package in x
    feat(minio): import updated minio package and check subscription plan
    feat(minio): import updated minio package and add tag on file upload

config/config.go
joremysh: updated and removed messages in PR description

jvallesm (approved):

PR #780, opened 2024-10-25
feat(collection): add task split

chuang8511:
Because

- it make pipeline easier to manipulate array

This commit

- add task split

### DIFF
--- a/pkg/component/generic/collection/v0/.compogen/bottom.mdx
+++ b/pkg/component/generic/collection/v0/.compogen/bottom.mdx
@@ -69,3 +69,32 @@ output:
     title: Object
     value: ${text-object.output.data}
 ```
+
+### Use `TASK_SPLIT` to split an array of strings into groups of a specified size
+```yaml
+# This pipeline splits an array of elements into groups of a specified size.
+# Examples:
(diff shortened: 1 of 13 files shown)


GeorgeWilliamStrong (commented):
From document perspective, LGTM

pinglin:
@chuang8511 Can we separate all tasks into `task_*` like what the Image operator has done? I found it's getting harder and harder to trace code in a single file across so many components and tasks. cc'd @donch1989 @jvallesm

jvallesm:
> @chuang8511 Can we separate all tasks into `task_*` like what the Image operator has done? I found it's getting harder and harder to trace code in a single file across so many components and tasks. cc'd @donch1989 @jvallesm

@pinglin do you mean in this case or as a convention?

To me it depends on the size of the `main.go` file. When we start a component and we have 1-2 (simple) tasks, I find it beneficial to have everything in a single file vs having several small files. As the component code grows, it makes sense to me to start dividing the logic into files. 

But it isn't a strongly-held opinion, I could do with a convention, too.

chuang8511:
Updated.

PR #791, opened 2024-10-30
feat(integration): identify supported OAuth integrations through global secrets

jvallesm:
Because

- We want to let CE users set up OAuth configuration for components.
- Until now, OAuth was considered as supported based on the component
  definition.

This commit

- Consider OAuth supported when both the component definition and the
  component global secrets have a complete OAuth configuration.


donch1989 (dismissed):

.github/workflows/integration-test.yml
```diff
@@ -71,14 +71,18 @@ jobs:
         uses: actions/checkout@v3
         with:
           repository: instill-ai/instill-core
+          ref: jvalles/ins-6570-pass-oauth-client-and-secret-ids-to-backend-and-frontend # TODO remove
```
jvallesm, line 74: Hi @donch1989 can you have a look at [this](https://github.com/instill-ai/instill-core/pull/1143), too? If it looks OK I'll merge that and remove this line

pkg/component/application/github/v0/main.go
jvallesm, line 1: ℹ️ @chuang8511 I extracted the OAuth support logic from the individual components as, at least for now, it will be the same code. Have a look at how I added it in the individual components `github` and `slack` (both in the component's `main` and on the `store` initialiser).

PR #793, opened 2024-10-30
feat: add task sync

chuang8511:
Because

- we want google drive component to be able to easily integrate into instill artifact
- we want to change the naming rule for the filename of task implementation file

This commit

- add task sync files
- delete operation file and create task files
- fix bugs in google drive
- add assertion helper to fail the test immediately when it should fail
  - In minimock function, quicktest won't fail the test immediately when the assertion is wrong.

Note
- QA result is in the thread of linear - QAed Test Cases


chuang8511:
QAed Test Cases in Linear ticket

chuang8511:
Not sure why Coverage / codecov keep failing..

PR #806, opened 2024-11-04
feat(component): add error handling for missing conversation

chuang8511:
Because

- it can improve UX when no error return in this case

This commit

- add test code
- clean the code by separating the tasks
- improve UX by return no error when no conversation


chuang8511:
Will need to do UAT again when UI is fixed.

PR #807, opened 2024-11-04
chore: fix pre-commit error

chuang8511:
Because

- there is error in pre-commit

This commit

- fix pre-commit config


chuang8511:
It seems the tag is ignored. So, it will probably get the same result like

image

I will close this and check in other branch again.

PR #819, opened 2024-11-06
feat(web): add input schema to improve web operator

chuang8511:
Because

- we want to filter urls

This commit

- add the filter function in web operator
- add termination system for context
- fix compogen bug

Note
- we update compogen to add blank line back in template. It is the reason why there are much more files changed.

### DIFF
--- a/pkg/component/operator/web/v0/config/tasks.json
+++ b/pkg/component/operator/web/v0/config/tasks.json
@@ -108,101 +108,37 @@
           "type": "integer"
         },
         "filter": {
-          "description": "A way to filter the URLs to crawl. If empty, all URLs will be crawled. Refer to the [Regex](#crawl-site-regex) and [Substring](#crawl-site-substring) for more details.",
+          "description": "Filtering based on [regular expression](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_expressions). The URL will be crawled if it matches either include-patterns or not match exclude-patterns. When both include-patterns and exclude-patterns are empty, all URLs will be crawled. It will process exclude-patterns first, then include-patterns. When exclude-patterns is not empty, only URLs that do not match exclude-patterns will be crawled. When include-patterns is not empty, only URLs that match include-patterns will be crawled.",
           "instillUIOrder": 5,
           "type": "object",
           "title": "Filter",
-          "oneOf": [
-            {
-              "properties": {
-                "filter-method": {
-                  "const": "Regex",
-                  "type": "string",
-                  "title": "Filter Method",
-                  "description": "Filtering based on regular expression.",
-                  "instillUIOrder": 0
-                },
-                "exclude-patterns": {
-                  "description": "When the URL is matched, the URL will not be crawled.",
-                  "instillAcceptFormats": [
-                    "array:string"
-                  ],
-                  "items": {
-                    "type": "string"
-                  },
-                  "instillUIOrder": 1,
-                  "type": "array",
-                  "title": "Exclude Patterns"
-                },
-                "include-patterns": {
-                  "description": "When the URL is matched, the URL will be crawled.",
-                  "instillAcceptFormats": [
-                    "array:string"
-                  ],
-                  "items": {
-                    "type": "string"
-                  },
-                  "instillUIOrder": 2,
-                  "type": "array",
-                  "title": "Include Patterns"
-                }
-              },
-              "required": [
-                "filter-method"
+          "properties": {
+            "exclude-patterns": {
+              "description": "When the URL is matched, the URL will not be crawled.",
+              "instillAcceptFormats": [
+                "array:string"
               ],
-              "title": "Regex",
-              "type": "object"
+              "items": {
+                "type": "string"
+              },
+              "instillUIOrder": 1,
+              "type": "array",
+              "title": "Exclude Patterns"
             },
-            {
-              "properties": {
-                "filter-method": {
-                  "const": "Substring",
-                  "type": "string",
-                  "title": "Filter Method",
-                  "description": "Filtering based on substring. Regex is more powerful than Substring. However, Substring is easier to use.",
-                  "instillUIOrder": 0
-                },
-                "exclude-substrings": {
-                  "description": "When the URL contains the substrings, the URL will not be crawled.",
-                  "instillAcceptFormats": [
-                    "array:string"
-                  ],
-                  "items": {
-                    "type": "string"
-                  },
-                  "instillUIOrder": 1,
-                  "type": "array",
-                  "title": "Exclude Substring"
-                },
-                "include-substrings": {
-                  "description": "When the URL contains the substrings, the URL will be crawled.",
-                  "instillAcceptFormats": [
-                    "array:string"
-                  ],
-                  "items": {
-                    "type": "string"
-                  },
-                  "instillUIOrder": 2,
-                  "type": "array",
-                  "title": "Include Substring"
-                },
-                "case-sensitive": {
-                  "description": "A flag indicating whether the substring matching is case-sensitive. When it is true, the matching is case-sensitive. When it is false, the matching is case-insensitive. The default value is false. For example, when it is case-sensitive, cat would only match 'cat' but not 'Cat' or 'CAT'. When cat is case-insensitive, on the other hand, would match 'cat', 'Cat', 'CAT', or any other variation of uppercase and lowercase letters.",
-                  "instillAcceptFormats": [
-                    "boolean"
-                  ],
-                  "instillUIOrder": 3,
-                  "title": "Case Sensitive",
-                  "type": "boolean"
-                }
-              },
-              "required": [
-                "filter-method"
+            "include-patterns": {
+              "description": "When the URL is matched, the URL will be crawled.",
+              "instillAcceptFormats": [
+                "array:string"
               ],
-              "title": "Substring",
-              "type": "object"
+              "items": {
+                "type": "string"
+              },
+              "instillUIOrder": 2,
+              "type": "array",
+              "title": "Include Patterns"
             }
-          ]
+          },
+          "required": []
         }
       },
       "required": [
--- a/pkg/component/operator/web/v0/config/tasks.json
+++ b/pkg/component/operator/web/v0/config/tasks.json
@@ -108,34 +108,28 @@
           "type": "integer"
         },
         "filter": {
-          "description": "Filtering based on [regular expression](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_expressions). The URL will be crawled if it matches either include-patterns or not match exclude-patterns. When both include-patterns and exclude-patterns are empty, all URLs will be crawled. It will process exclude-patterns first, then include-patterns. When exclude-patterns is not empty, only URLs that do not match exclude-patterns will be crawled. When include-patterns is not empty, only URLs that match include-patterns will be crawled.",
+          "description": "Filtering based on [regular expression](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_expressions). The URL will be crawled if it matches either include-pattern or not match exclude-pattern. When both include-pattern and exclude-pattern are empty, all URLs will be crawled. It will process exclude-pattern first, then include-pattern. When exclude-pattern is not empty, only URLs that do not match exclude-pattern will be crawled. When include-pattern is not empty, only URLs that match include-pattern will be crawled.",
           "instillUIOrder": 5,
           "type": "object",
           "title": "Filter",
           "properties": {
-            "exclude-patterns": {
+            "exclude-pattern": {
               "description": "When the URL is matched, the URL will not be crawled.",
               "instillAcceptFormats": [
-                "array:string"
+                "string"
               ],
-              "items": {
-                "type": "string"
-              },
               "instillUIOrder": 1,
-              "type": "array",
-              "title": "Exclude Patterns"
+              "type": "string",
+              "title": "Exclude Pattern"
             },
-            "include-patterns": {
+            "include-pattern": {
               "description": "When the URL is matched, the URL will be crawled.",
               "instillAcceptFormats": [
-                "array:string"
+                "string"
               ],
-              "items": {
-                "type": "string"
-              },
               "instillUIOrder": 2,
-              "type": "array",
-              "title": "Include Patterns"
+              "title": "Include Pattern",
+              "type": "string"
             }
           },
           "required": []
--- a/pkg/component/operator/web/v0/config/tasks.json
+++ b/pkg/component/operator/web/v0/config/tasks.json
@@ -106,6 +106,103 @@
           "minimum": 0,
           "title": "Max Depth",
           "type": "integer"
+        },
+        "filter": {
+          "description": "A way to filter the URLs to crawl. If empty, all URLs will be crawled. Refer to the [Regex](#crawl-site-regex) and [Substring](#crawl-site-substring) for more details.",
+          "instillUIOrder": 5,
+          "type": "object",
+          "title": "Filter",
+          "oneOf": [
+            {
+              "properties": {
+                "filter-method": {
+                  "const": "Regex",
+                  "type": "string",
+                  "title": "Filter Method",
+                  "description": "Filtering based on regular expression.",
+                  "instillUIOrder": 0
+                },
+                "exclude-patterns": {
+                  "description": "When the URL is matched, the URL will not be crawled.",
+                  "instillAcceptFormats": [
+                    "array:string"
+                  ],
+                  "items": {
+                    "type": "string"
+                  },
+                  "instillUIOrder": 1,
+                  "type": "array",
+                  "title": "Exclude Patterns"
+                },
+                "include-patterns": {
+                  "description": "When the URL is matched, the URL will be crawled.",
+                  "instillAcceptFormats": [
+                    "array:string"
+                  ],
+                  "items": {
+                    "type": "string"
+                  },
+                  "instillUIOrder": 2,
+                  "type": "array",
+                  "title": "Include Patterns"
+                }
+              },
+              "required": [
+                "filter-method"
+              ],
+              "title": "Regex",
+              "type": "object"
+            },
+            {
+              "properties": {
+                "filter-method": {
+                  "const": "Substring",
+                  "type": "string",
+                  "title": "Filter Method",
+                  "description": "Filtering based on substring. Regex is more powerful than Substring. However, Substring is easier to use.",
+                  "instillUIOrder": 0
+                },
+                "exclude-substrings": {
+                  "description": "When the URL contains the substrings, the URL will not be crawled.",
+                  "instillAcceptFormats": [
+                    "array:string"
+                  ],
+                  "items": {
+                    "type": "string"
+                  },
+                  "instillUIOrder": 1,
+                  "type": "array",
+                  "title": "Exclude Substring"
+                },
+                "include-substrings": {
+                  "description": "When the URL contains the substrings, the URL will be crawled.",
+                  "instillAcceptFormats": [
+                    "array:string"
+                  ],
+                  "items": {
+                    "type": "string"
+                  },
+                  "instillUIOrder": 2,
+                  "type": "array",
+                  "title": "Include Substring"
+                },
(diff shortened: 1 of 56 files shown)


pkg/component/operator/web/v0/crawl_website.go
```diff
@@ -39,6 +39,18 @@ type CrawlWebsiteInput struct {
 	Timeout int `json:"timeout"`
 	// MaxDepth: The maximum depth of the pages to scrape.
 	MaxDepth int `json:"max-depth"`
+	// Filter: The filter to filter the URLs to crawl.
+	Filter Filter `json:"filter"`
+}
+
+// Filter defines the filter of the crawl website task
+type Filter struct {
```
jvallesm, line 47: nit: can be unexported

pkg/component/operator/web/v0/config/tasks.json
```diff
@@ -108,101 +108,37 @@
...
-              "required": [
-                "filter-method"
+              "instillUIOrder": 1,
+              "type": "array",
+              "title": "Exclude Patterns"
+            },
+            "include-patterns": {
+              "description": "When the URL is matched, the URL will be crawled.",
+              "instillAcceptFormats": [
+                "array:string"
```
jvallesm, line 131: If regexp has an OR filter (as  in the examples you provided), what's the use case for passing several expressions?

pkg/component/operator/web/v0/crawl_website.go
```diff
@@ -39,6 +39,18 @@ type CrawlWebsiteInput struct {
...
+	// ExcludePatterns: The patterns to exclude the URLs to crawl.
+	ExcludePatterns []string `json:"exclude-patterns"`
+	// IncludePatterns: The patterns to include the URLs to crawl.
+	IncludePatterns []string `json:"include-patterns"`
+	// ExcludeSubstrings: The substrings to exclude the URLs to crawl.
+	ExcludeSubstrings []string `json:"exclude-substrings"`
```
jvallesm, line 53: This isn't defined

pkg/component/operator/web/v0/crawl_website.go
```diff
@@ -105,6 +117,10 @@ func (e *execution) CrawlWebsite(input *structpb.Struct) (*structpb.Struct, erro
 
 		link := e.Attr("href")
 
+		if !targetLink(link, inputStruct.Filter) {
```
jvallesm: nit: this could be more efficient by compiling the expressions in the filter only once, then using them for all the links in the execution. I don't think this will ever be a bottleneck but I wanted to highlight that regular expressions aren't particularly cheap, computationally speaking.

pkg/component/operator/web/v0/helper.go
```diff
@@ -65,3 +66,29 @@ func getRemovedTagsHTML[T scrapeInput](doc *goquery.Document, input T) string {
 
 	return combinedHTML
 }
+
+// targetLink filters the URL based on the filter
+func targetLink(link string, filter Filter) bool {
+
+	if len(filter.ExcludePatterns) == 0 && len(filter.IncludePatterns) == 0 {
```
jvallesm, line 73: This condition is redundant with the following code, as range effectively works as a length check

pkg/component/operator/web/v0/crawl_website.go
```diff
@@ -198,6 +209,33 @@ func (e *execution) CrawlWebsite(input *structpb.Struct) (*structpb.Struct, erro
...
+				// If no new pages for 2 seconds, cancel the context
+				cancel()
+				return
+			case <-ctx.Done():
+				return
+			}
+		}
+	}()
+
 	<-ctx.Done()
```
jvallesm, line 249: Is this redundant with the previous `<-ctx.Done()`?

pkg/component/operator/web/v0/crawl_website.go
```diff
@@ -198,6 +209,33 @@ func (e *execution) CrawlWebsite(input *structpb.Struct) (*structpb.Struct, erro
...
+
+		for {
+			select {
+			case <-pageUpdateCh:
+				// Reset the timer whenever we get a new page
+				if !inactivityTimer.Stop() {
+					select {
+					case <-inactivityTimer.C:
+					default:
+					}
```
jvallesm, line 236: Again, why a select statement with a default instead of reading the channel? Can't you just read the timer? Also, why do we need to `Stop` instead of just `Reset`ting the timer?

pkg/component/operator/web/v0/crawl_website.go
```diff
@@ -177,6 +182,12 @@ func (e *execution) CrawlWebsite(input *structpb.Struct) (*structpb.Struct, erro
 		if len(output.Pages) < inputStruct.MaxK {
 			output.Pages = append(output.Pages, page)
 
+			// Signal that we've added a new page
+			select {
+			case pageUpdateCh <- struct{}{}:
+			default:
```
jvallesm, line 198: Why do you need a select with default here instead of sending a value to the channel directly?

pkg/component/operator/web/v0/crawl_website.go
```diff
@@ -148,6 +148,11 @@ func (e *execution) CrawlWebsite(input *structpb.Struct) (*structpb.Struct, erro
 		r.Headers.Set("User-Agent", randomString())
 	})
 
+	// colly.Wait() does not terminate the program. So, we need a system to terminate the program when there is no collector.
+	// We use a channel to notify the main goroutine that a new page has been scraped.
+	// When there is no new page for 2 seconds, we cancel the context.
```
jvallesm, line 163: Silly question: is there a way to know there aren't remaining URLs to scrape (e.g. cancelling the context after `Wait` finishes)? Even if there's a sweet spot, waiting for any amount of time is always a compromise between being inefficient (waiting when we could return) and risking data loss (there's new data but it's taking longer than the threshold).

chuang8511: Originally, I expected colly should finish the job when there aren't remaining URLs.

But, I could not find a way to close colly when there is no callback jobs.
So, I decided to use this way to deal with the case that the output length hasn't achieved the max-k but there is no URLs.

pkg/component/operator/web/v0/crawl_website.go
chuang8511: We need this to block the main execution.

chuang8511:
@jvallesm 
Thanks for reviewing. 
I modified and replied all of them. 
Please take a look when you have time!

pkg/component/operator/web/v0/config/tasks.json
```diff
@@ -106,6 +106,103 @@
...
+        "filter": {
+          "description": "A way to filter the URLs to crawl. If empty, all URLs will be crawled. Refer to the [Regex](#crawl-site-regex) and [Substring](#crawl-site-substring) for more details.",
+          "instillUIOrder": 5,
+          "type": "object",
+          "title": "Filter",
+          "oneOf": [
+            {
+              "properties": {
+                "filter-method": {
+                  "const": "Regex",
```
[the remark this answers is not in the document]
jvallesm: Resolving this comment as outdated, but still relevant for future development!

pkg/component/operator/web/v0/crawl_website.go
jvallesm: Yeah but you can skip the `go func()` in the previous block and run the `for` / `select` in the main thread, right? Otherwise it looks like you're dispatching work in order to wait for it immediately.

pkg/component/operator/web/v0/crawl_website.go
jvallesm: Something that isn't clear to me is what you mean by 
> does not terminate the program. So, we need a system to terminate the program when there is no collector.

I'm not familiar with `colly`, so what I'm proposing might not have sense (in which case, I'm just trying to understand the code). Colly is a popular enough project, so my intuition tells me there has to be a simpler way to finish the execution when all the websites have been crawled. Also I'm assuming that `c.Wait()` does work and that it halts the execution until all the sites have been scraped.

From the assumption I just mentioned, what I see you're trying to accomplish is waiting until the first occurrence between:
- All the sites have been scraped. `c.Wait()` will signal this.
- We've reached the 2 minute timeout. `ctx.Done()` will signal this.
- We've reached the limit of pages. `ctx.Done()` will also signal this, as we cancel the context when we find this condition.

Therefore, we need to wait for either `c.Wait()` or `ctx.Done()`. I think we can avoid timing the duration between 2 completed pages and select between `ctx.Done()` and a channel that is closed after `c.Wait()` ends:

```go
	scrapeDone := make(chan struct{})

	go func() {
		defer close(scrapeDone)
		_ = c.Visit(inputStruct.URL)
		c.Wait()
	}()

	select {
	case <- scrapeDone:
		// No more pages to scrape.
	case <-ctx.Done():
		// Timeout or page limit reached
	}
```

chuang8511: > All the sites have been scraped. `c.Wait()` will signal this.

Yes! It is the problem. I expect `c.Wait()` will finish when there is no more callback (job).
However, when I did end-to-end test, I noticed it did not close this go-routine. 
If `c.Wait()` can close it, we won't need another `chan`(If I understand it correctly).

pkg/component/operator/web/v0/crawl_website.go
```diff
@@ -73,7 +106,22 @@ func (e *execution) CrawlWebsite(input *structpb.Struct) (*structpb.Struct, erro
...
+	err = inputStruct.Filter.compile()
+	if err != nil {
+		return nil, fmt.Errorf("compiling filter: %v", err)
+	}
+
+	output := ScrapeWebsiteOutput{
+		Pages: []PageInfo{},
+	}
+
+	if !targetLink(inputStruct.URL, inputStruct.Filter) {
```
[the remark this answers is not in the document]
chuang8511: Thanks! I will consider it!

pkg/component/operator/web/v0/crawl_website.go
chuang8511: Oh, do you mean we actually don't need another go-routine for this block?
So, `<-ctx.Done()` in the main thread is not necessary.

```
		inactivityTimer := time.NewTimer(2 * time.Second)
		defer inactivityTimer.Stop()

		for {
			select {
			case <-pageUpdateCh:
				inactivityTimer.Reset(2 * time.Second)
			// If no new pages for 2 seconds, cancel the context
			case <-inactivityTimer.C:
				cancel()
				return
			// If the context is done, we should return
			case <-ctx.Done():
				return
			}
		}
```

pkg/component/operator/web/v0/crawl_website.go
jvallesm: > I noticed it did not close this go-routine.

I don't fully understand this. How were you checking the goroutine had ended? The moment you use `go func`, the main thread stops caring about the function you're calling, so it will never know when `c.Wait` finished. That's why in  my snippet I added a channel that's closed at the end of the spawned function. By listening to that channel, the main thread is signaled about the wait group being unlocked.

chuang8511: Oh, I think I get the point. I will try your proposal.

jvallesm: Let me know how it goes!

chuang8511: It doesn't work. 😢 .
I guess the only way  to finish `c.Wait()` is `r.Abort()` or cancel the ctx.

However, we cannot abort it if the the output page count has not achieved maxK.

pkg/component/operator/web/v0/crawl_website.go
chuang8511: > depending on the thread about c.Wait()

It does not work, so I modified the code [here](https://github.com/instill-ai/pipeline-backend/pull/819/commits/4a9055a3dfd9a03ec24da8527428bebf70f06a74).

chuang8511:
@jvallesm I modified it again. Please take a look! Thank you!

pkg/component/operator/web/v0/crawl_website.go
jvallesm: 🤔 That is weird, as I mentioned the documentation indicates that `Wait` should signal the end of the scraping and the project is popular enough as to discard an issue here.

I actually [modified their parallel example](https://gist.github.com/jvallesm/b542f1b377863deafeecb8b434f996ed) (I also included URL filtering if you want to check it out) and, indeed, the strategy I suggested works.

Therefore, this claim is incorrect, there must be something else preventing the collector from finishing.
>  I guess the only way to finish c.Wait() is r.Abort() or cancel the ctx.

chuang8511: I am not sure if I am right. But, I imagine like the flow blow.

When `OnHTML` is triggered, it collects the to be scraped URLs in a queue.
The workers will scrape the URLs in the queue in the sequence.

And, `OnHTML` will be triggered in a big count in the beginning, which makes the the queue contains lots of URLs that need workers to process.

So, that is the reason why the workers do not close.

But, it comes to another question. Why do we need to close the worker if they haven't finished?
It is because those remaining pages aren't probably the target pages because the crawler failed to get data by the timeout setting.

So, it seems the remaining URLs are the URLs that the workers need to process that to wait for the timeout and cannot get any data.

If this assumption is correct, we will need a timer to close the worker to avoid the worker to wait for all URLs' timeout.

The [sample](https://gist.github.com/chuang8511/4f721be9b3c9de648d725162ccefffce) is just printing out the `htmlcount` & `output count`.

jvallesm: But in that scenario, we're not talking about the process having finished, it's just that each new visit takes longer and we want to set a timeout because that _probably_ means these results aren't interesting. In that case, indeed `Wait()` won't work, but then we're not listening to the end of the process. Then we can wait for the total timeout or set a timeout between requests, which I don't love but I see the point about.

In your example (and here), should we cancel the context when the `MaxK` has been reached? Otherwise requests are skipped but the crawling goes on. I don't know if this is intentional.

Also, isn't it simpler to use channels here than mutexes?

chuang8511: > should we cancel the context when the MaxK has been reached?

Yes, we should. When the output reaches the count the users want, we can return the result to the users. So, we should cancel the context and stop requesting and crawling. 

> Also, isn't it simpler to use channels here than mutexes?

I am not quite sure what it specifically means in which part. We do use channels to signal the stop process here. right?

jvallesm: > > should we cancel the context when the MaxK has been reached?
>
> Yes, we should. When the output reaches the count the users want, we can return the result to the users. So, we should cancel the context and stop requesting and crawling.

I mentioned that because your sample didn't do that, so the code kept crawling after the max results had been collected and it ran until the timeout was reached. If you cancel the context (or close the `scrapeDone` channel), the program doesn't reach the timeout. Adding this, we can check the 3 scenarios work (in your sample, we'd need to port that to the component code):

- No more pages to scrape: `c.Wait()` returns and the goroutine closes `scrapeDone`. Program finishes before the timeout.
- Max pages scraped: `c.OnResponse` cancels the context / closes `scrapeDone`. Program finishes before the timeout.
- Max pages haven't been collected before the timeout. Context is canceled and program finishes at the timeout.

This is the modification I made:
```go
	c.OnResponse(func(r *colly.Response) {
		log.Println(r.Request.URL.String())
		mu.Lock()
		defer mu.Unlock()
		responsecount++
		fmt.Println("responsecount", responsecount)
		output = append(output, r.Request.URL.String())
		if len(output) >= 10 {
			close(scrapeDone) // cancel()
			return
		}
		fmt.Println(r.StatusCode)
	})
```

jvallesm: > > Also, isn't it simpler to use channels here than mutexes?
>
> I am not quite sure what it specifically means in which part. We do use channels to signal the stop process here. right?

What I  mean is that you're defining variables outside of the goroutines and using mutexes to make sure you modify them atomically. To me a much more natural pattern is using channels to communicate the information outside of the routines. There's [this article](https://go.dev/wiki/MutexOrChannel) and also the [go proverb](https://go-proverbs.github.io/):

> [Channels orchestrate; mutexes serialize.](https://www.youtube.com/watch?v=PAAkCSZUG1c&t=4m20s).

chuang8511: > What I mean is that you're defining variables outside of the goroutines and using mutexes to make sure you modify them atomically.

I got what you mean here. You mean the output pages. right? 
I will address this point in other PR and take care of this part in the future development.


> Adding this, we can check the 3 scenarios work (in your sample, we'd need to port that to the component code):

Let's back to the current code(PR) not sample one. 
How about I do the following thing.

I will add back the `scrapeDone` channel and remain the "2 second part" in the codebase. 
So, we can deal with 4 scenarios.

1. No more pages to scrape: c.Wait() returns and the goroutine closes scrapeDone. Program finishes before the timeout.
2. Max pages scraped: c.OnResponse cancels the context / closes scrapeDone. Program finishes before the timeout.
3. Max pages haven't been collected before the timeout. Context is canceled and program finishes at the timeout.
4. Max pages have not been collected before the timeout. Context is canceled and the program finishes when there are no more validate data in 2 seconds.

And, why we need the 4th one is because what I mentioned [here](https://github.com/instill-ai/pipeline-backend/pull/819#discussion_r1843668582).

jvallesm: Agreed, make sure you document the 4th in the code 🤝 I [updated the sample](https://gist.github.com/jvallesm/b542f1b377863deafeecb8b434f996ed) to use channels.

chuang8511: Added [it](https://github.com/instill-ai/pipeline-backend/pull/819/commits/1d23911395865ae5abe5cde048028315a5b5db8a) back.

pkg/component/operator/web/v0/crawl_website.go
```diff
@@ -177,12 +242,41 @@ func (e *execution) CrawlWebsite(input *structpb.Struct) (*structpb.Struct, erro
...
+	// To avoid to wait for 2 minutes, we use a timer to check if there is a new page.
+	// If there is no new page, we cancel the context.
+	inactivityTimer := time.NewTimer(2 * time.Second)
+	defer inactivityTimer.Stop()
+
+	// There are 4 cases to finish the program:
+	// 1. No more pages to scrape: c.Wait() returns and the goroutine closes scrapeDone. Program finishes before the timeout.
+	// 2. Max pages scraped: c.OnResponse cancels the context / closes scrapeDone. Program finishes before the timeout.
+	// 3. Max pages haven't been collected before the timeout. Context is canceled and program finishes at the timeout.
+	// 4. Max pages haven't been collected before the timeout. Context is canceled and the program finishes when there are no more validate data in 2 seconds.
```
jvallesm, line 261:
```suggestion
	// 4. Max pages haven't been collected but the delay between completed requests has increased over 2 seconds. We consider this data invalid because (insert reason here), so we avoid waiting for the timeout.
```

pkg/component/operator/web/v0/crawl_website.go
```diff
@@ -177,12 +242,41 @@ func (e *execution) CrawlWebsite(input *structpb.Struct) (*structpb.Struct, erro
...
+	// We use 4. to avoid the program waiting for 2 minutes to close all URLs that will wait for over timeout.
```
jvallesm, line 262: This is still not super clear to me 😅

pkg/component/operator/web/v0/crawl_website.go
```diff
@@ -104,12 +152,23 @@ func (e *execution) CrawlWebsite(input *structpb.Struct) (*structpb.Struct, erro
...
+		}
 
-		if util.InSlice(pageLinks, link) {
+		parsedURL, err := url.Parse(link)
+		if err != nil {
 			return
 		}
 
-		pageLinks = append(pageLinks, link)
+		requestURL := stripQueryAndTrailingSlash(parsedURL)
```
jvallesm, line 165: Why did you add this bit? Is it related with the URL filter?

jvallesm (commented):
I added some post-merge comments that aren't a blocker to move forward with deploying this.

PR #830, opened 2024-11-10
refactor(component,collection): refactor codebase and fix task

pinglin:
Because

- the collection implementation and `README.md` have shown inconsistent naming and code styles
- the current implementation is relatively limited to only `array`, in which we'd like to extend to also `object` and `primitive` type

This commit

- uses coherent `data` for input and output to replace with `set`
- improve task descriptions
- improve task capability to cope with `array`, `object`, and `primitive` (if applicable)


pkg/component/generic/collection/v0/io.go
```diff
@@ -5,60 +5,70 @@ import (
...
-	Set []format.Value `instill:"set"`
+	Data format.Value `instill:"data"`
 }
 
 type splitInput struct {
-	Array     []format.Value `instill:"array"`
-	GroupSize int            `instill:"group-size"`
+	Data      format.Value `instill:"data"`
+	Size      int          `instill:"size"`
+	Delimiter string       `instill:"delimiter"`
```
donch1989, line 53: I believe this delimiter is unused and can be removed.

donch1989:
Let’s hold off on merging this PR. I’ll check tomorrow if we can create a migration script for it.

donch1989:
OK.. we don't have a easy way to migrate the TASK_ASSIGN, let's merge it.

PR #834, opened 2024-11-12
feat(vdp): integrate blob storage to vdp

chuang8511:
Because

- we want to read data directly from minio
- we want to upload output data to minio directly and return download url

This commit

- upload output data to minio
- read data from minio
- inject artifact clients to service and worker for the future usage

Note
- We will need to add test code for temporal worker

### DIFF
--- a/pkg/data/client/client.go
+++ b/pkg/data/client/client.go
@@ -1,0 +1,55 @@
+package client
+
+import (
+	"context"
+	"fmt"
+
+	"go.uber.org/zap"
+	"google.golang.org/grpc"
+
+	"github.com/instill-ai/pipeline-backend/config"
+	"github.com/instill-ai/pipeline-backend/pkg/external"
+
+	artifactpb "github.com/instill-ai/protogen-go/artifact/artifact/v1alpha"
+	miniox "github.com/instill-ai/x/minio"
+)
+
+var (
+	instance *Clients
+)
+
+// BlobStorage is an interface for fetching files from blob storage
+type BlobStorage interface {
+	// GetFile fetches a file from blob storage
+	GetFile(ctx context.Context, bucketName, objectPath string) (data []byte, contentType string, err error)
+}
+
+// Clients is a struct that holds all the clients
+type Clients struct {
+	// ArtifactPublicServiceClient can fetch metadata from artifact service
+	ArtifactPrivateServiceClient artifactpb.ArtifactPrivateServiceClient
+	// GRPCConn is the connection to the artifact private service
+	GRPCConn *grpc.ClientConn
+	// BlobStorageClient can fetch files from blob storage
+	BlobStorageClient BlobStorage
+}
+
+// GetClients returns an instance of Clients
+func GetClients(ctx context.Context, logger *zap.Logger) (*Clients, error) {
+	clients := Clients{}
+	artifactPrivateServiceClient, privConn := external.InitArtifactPrivateServiceClient(ctx)
+	if artifactPrivateServiceClient == nil {
+		return nil, fmt.Errorf("artifact private service client is nil")
+	}
+	clients.GRPCConn = privConn
+	clients.ArtifactPrivateServiceClient = artifactPrivateServiceClient
+
+	minioClient, err := miniox.NewMinioClient(ctx, &config.Config.Minio, logger)
+	if err != nil {
+		return nil, err
+	}
+	clients.BlobStorageClient = minioClient
+	instance = &clients
+
+	return instance, nil
+}
--- a/pkg/data/file.go
+++ b/pkg/data/file.go
@@ -14,10 +14,10 @@ import (
 	"github.com/gofrs/uuid"
 	"google.golang.org/protobuf/types/known/structpb"
 
-	"github.com/instill-ai/pipeline-backend/pkg/data/client"
 	"github.com/instill-ai/pipeline-backend/pkg/data/format"
 	"github.com/instill-ai/pipeline-backend/pkg/data/path"
 	"github.com/instill-ai/pipeline-backend/pkg/logger"
+	"github.com/instill-ai/pipeline-backend/pkg/repository"
 
 	artifactpb "github.com/instill-ai/protogen-go/artifact/artifact/v1alpha"
 )
@@ -78,11 +78,11 @@ func convertURLToBytes(url string) (b []byte, contentType string, filename strin
 func fetchFileFromBlobStorage(urlUID uuid.UUID) (b []byte, contentType string, filename string, err error) {
 	ctx := context.Background()
 	logger, _ := logger.GetZapLogger(ctx)
-	clients, err := client.GetClients(ctx, logger)
+	clients, err := repository.GetClients(ctx, logger)
 	if err != nil {
 		return nil, "", "", err
 	}
-	defer clients.GRPCConn.Close()
+	defer clients.Close()
 
 	artifactClient := clients.ArtifactPrivateServiceClient
 	objectURLRes, err := artifactClient.GetObjectURL(ctx, &artifactpb.GetObjectURLRequest{
--- a/pkg/data/file.go
+++ b/pkg/data/file.go
@@ -56,6 +56,7 @@ func convertURLToBytes(url string) (b []byte, contentType string, filename strin
 	if strings.HasPrefix(url, "data:") {
 		return convertDataURIToBytes(url)
 	}
+	// TODO: judge URL and read data from minio, read data implementation will be in x repo.
 
 	client := resty.New().SetRetryCount(3)
 	resp, err := client.R().Get(url)
--- a/pkg/data/file.go
+++ b/pkg/data/file.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/gabriel-vasile/mimetype"
 	"github.com/go-resty/resty/v2"
+	"github.com/gofrs/uuid"
 	"google.golang.org/protobuf/types/known/structpb"
 
 	"github.com/instill-ai/pipeline-backend/pkg/data/client"
@@ -35,8 +36,8 @@ var fileGetters = map[string]func(*fileData) (format.Value, error){
 	"base64":       func(f *fileData) (format.Value, error) { return f.Base64() },
 }
 
-// Pattern matches: v1alpha/namespaces/{namespace}/blob-urls/{uid}
-var minioURLPattern = regexp.MustCompile(`/v1alpha/namespaces/([^/]+)/blob-urls/([^/]+)$`)
+// Pattern matches: https://{domain}/v1alpha/namespaces/{namespace}/blob-urls/{uid}
+var minioURLPattern = regexp.MustCompile(`https?://[^/]+/v1alpha/namespaces/[^/]+/blob-urls/([^/]+)$`)
 
 type fileData struct {
 	raw         []byte
@@ -66,15 +67,15 @@ func convertURLToBytes(url string) (b []byte, contentType string, filename strin
 		return convertDataURIToBytes(url)
 	}
 	if matches := minioURLPattern.FindStringSubmatch(url); matches != nil {
-		if len(matches) < 3 {
+		if len(matches) < 2 {
 			return nil, "", "", fmt.Errorf("invalid blob storage url: %s", url)
 		}
-		return fetchFileFromBlobStorage(matches[2])
+		return fetchFileFromBlobStorage(uuid.FromStringOrNil(matches[1]))
 	}
 	return fetchExternalURL(url)
 }
 
-func fetchFileFromBlobStorage(urlUID string) (b []byte, contentType string, filename string, err error) {
+func fetchFileFromBlobStorage(urlUID uuid.UUID) (b []byte, contentType string, filename string, err error) {
 	ctx := context.Background()
 	logger, _ := logger.GetZapLogger(ctx)
 	clients, err := client.GetClients(ctx, logger)
@@ -85,7 +86,7 @@ func fetchFileFromBlobStorage(urlUID string) (b []byte, contentType string, file
 
 	artifactClient := clients.ArtifactPrivateServiceClient
 	objectURLRes, err := artifactClient.GetObjectURL(ctx, &artifactpb.GetObjectURLRequest{
-		Uid: urlUID,
+		Uid: urlUID.String(),
 	})
 	if err != nil {
 		return nil, "", "", err
--- a/pkg/data/file.go
+++ b/pkg/data/file.go
@@ -2,8 +2,10 @@ package data
 
 import (
 	"bytes"
+	"context"
 	"fmt"
 	"mime"
+	"regexp"
 	"strings"
 	"unicode/utf8"
 
@@ -11,8 +13,12 @@ import (
 	"github.com/go-resty/resty/v2"
 	"google.golang.org/protobuf/types/known/structpb"
 
+	"github.com/instill-ai/pipeline-backend/pkg/data/client"
 	"github.com/instill-ai/pipeline-backend/pkg/data/format"
 	"github.com/instill-ai/pipeline-backend/pkg/data/path"
+	"github.com/instill-ai/pipeline-backend/pkg/logger"
+
+	artifactpb "github.com/instill-ai/protogen-go/artifact/artifact/v1alpha"
 )
 
 const (
@@ -28,6 +34,9 @@ var fileGetters = map[string]func(*fileData) (format.Value, error){
 	"data-uri":     func(f *fileData) (format.Value, error) { return f.DataURI() },
 	"base64":       func(f *fileData) (format.Value, error) { return f.Base64() },
 }
+
+// Pattern matches: v1alpha/namespaces/{namespace}/blob-urls/{uid}
+var minioURLPattern = regexp.MustCompile(`/v1alpha/namespaces/([^/]+)/blob-urls/([^/]+)$`)
 
 type fileData struct {
 	raw         []byte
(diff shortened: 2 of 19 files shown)


pkg/worker/workflow.go
```diff
@@ -604,6 +617,108 @@ func (w *worker) OutputActivity(ctx context.Context, param *ComponentActivityPar
...
+	default:
+		return v
+	}
+}
+
+func (w *worker) uploadBlobDataAndGetDownloadURL(ctx context.Context, param *ComponentActivityParam, value *format.File) (string, error) {
+	artifactClient := w.artifactPublicServiceClient
+	ns := param.Namespace
+
+	jsonBytes, err := json.Marshal(param.SystemVariables)
```
donch1989, line 654: Marshalling and unmarshalling it back again seems redundant.

pkg/worker/workflow.go
```diff
@@ -604,6 +617,108 @@ func (w *worker) OutputActivity(ctx context.Context, param *ComponentActivityPar
 	logger.Info("OutputActivity completed")
 	return nil
 }
+func (w *worker) uploadFileAndReplaceURL(ctx context.Context, param *ComponentActivityParam, value *format.Value) format.Value {
```
donch1989, line 620:
```suggestion
func (w *worker) uploadFileAndReplaceWithURL(ctx context.Context, param *ComponentActivityParam, value *format.Value) format.Value {
```

pkg/data/file.go
```diff
@@ -29,6 +35,9 @@ var fileGetters = map[string]func(*fileData) (format.Value, error){
 	"base64":       func(f *fileData) (format.Value, error) { return f.Base64() },
 }
 
+// Pattern matches: v1alpha/namespaces/{namespace}/blob-urls/{uid}
+var minioURLPattern = regexp.MustCompile(`/v1alpha/namespaces/([^/]+)/blob-urls/([^/]+)$`)
```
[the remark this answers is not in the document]
chuang8511: It can deal with the URL with domain name. I will make it clear in comment.

pkg/worker/workflow.go
```diff
@@ -44,6 +48,8 @@ type TriggerPipelineWorkflowParam struct {
 	Mode            mgmtpb.Mode
 	TriggerFromAPI  bool
 	WorkerUID       uuid.UUID
+	// Namespace is the namespace of the requester of the pipeline.
+	Namespace resource.Namespace
```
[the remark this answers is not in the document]
chuang8511: I realised that we only need nsID, so we actually don't need to inject this Namespace into this struct.
But, I am confused which nsID we should use. There are User / Owner / Requester.
I think we should use `PipelineUserUID`. right?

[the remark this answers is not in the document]
chuang8511: Ah, no, I was wrong. We still need to add Ns.ID.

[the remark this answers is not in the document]
chuang8511: And, after checking the upstream data, I realised that owner is from `ns`. I merged it and will push the change.

chuang8511:
@donch1989 Thanks for reviewing. I modified all.

pkg/worker/workflow.go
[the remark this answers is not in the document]
jvallesm: > There are User / Owner / Requester

- User: the authenticated user making the request
- Requester: the namespace who's making the request
- Owner: the owner of the pipeline

The trigger data you want to upload belongs to the `Requester`, same as e.g. the run logs or the credit consumption.

pkg/data/client/client.go
jvallesm, line 1: Should this be in `data`? I'm not very familiar with this package but it seems to hold logic without any clients. If we're talking about storage, shouldn't `repository` be more adequate?

pkg/service/main.go
```diff
@@ -92,52 +93,67 @@ type TriggerResult struct {
...
+	mgmtPrivateServiceClient     mgmtpb.MgmtPrivateServiceClient
+	aclClient                    acl.ACLClientInterface
+	converter                    Converter
+	minioClient                  miniox.MinioI
+	memory                       memory.MemoryStore
+	log                          *zap.Logger
+	workerUID                    uuid.UUID
+	retentionHandler             MetadataRetentionHandler
+	artifactPublicServiceClient  artifactpb.ArtifactPublicServiceClient
+	artifactPrivateServiceClient artifactpb.ArtifactPrivateServiceClient
```
jvallesm, line 114: ❓ Is this one used?

pkg/data/client/client.go
```diff
@@ -0,0 +1,55 @@
...
+	}
+	clients.GRPCConn = privConn
+	clients.ArtifactPrivateServiceClient = artifactPrivateServiceClient
+
+	minioClient, err := miniox.NewMinioClient(ctx, &config.Config.Minio, logger)
+	if err != nil {
+		return nil, err
+	}
+	clients.BlobStorageClient = minioClient
+	instance = &clients
```
jvallesm, line 52: What's the benefit of storing the clients in a package-scoped unexported variable?

pkg/data/file.go
```diff
@@ -56,7 +66,55 @@ func convertURLToBytes(url string) (b []byte, contentType string, filename strin
...
+	if matches := minioURLPattern.FindStringSubmatch(url); matches != nil {
+		if len(matches) < 2 {
+			return nil, "", "", fmt.Errorf("invalid blob storage url: %s", url)
+		}
+		return fetchFileFromBlobStorage(uuid.FromStringOrNil(matches[1]))
+	}
+	return fetchExternalURL(url)
+}
+
+func fetchFileFromBlobStorage(urlUID uuid.UUID) (b []byte, contentType string, filename string, err error) {
```
jvallesm, line 78: This is implemented as a pure function (without any receiver) but, underneath, it initialises (or leverages some initialised) clients and dependencies. This is misleading and I'd even say unidiomatic. A much more natural way / more aligned with our codebase would be having a `BlobFetcher` structure which holds the dependencies (artifact client, MinIO client, even an HTTP client). These should be initialised on the `main` package and injected to this struct.

jvallesm (commented):
I think the way the blob fetcher has been added to the codebase isn't very clean nor aligned with the current structure, though I didn't find any issues  with the functionality.

pkg/worker/workflow.go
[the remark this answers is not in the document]
donch1989: Hmm, this raises an important question: should this files be placed under the owner’s namespace or the requester’s namespace?

pkg/data/client/client.go
donch1989: Agree that we can put if under repository.

pkg/data/file.go
donch1989: I suspect this issue is caused by the current data package implementation.

The `fetchFileFromBlobStorage` function is invoked by `NewFileFromURL()`, but `NewFileFromURL()` doesn’t inject any client since it might be used in various places.

@jvallesm, are you suggesting that we should inject the client through `NewFileFromURL()`? If so, should we also inject the Resty client from the outside? This would make the function signature something like:
`NewFileFromURL(url, fetcher)`.

pkg/data/file.go
```diff
@@ -56,7 +66,55 @@ func convertURLToBytes(url string) (b []byte, contentType string, filename strin
...
+	ctx := context.Background()
+	logger, _ := logger.GetZapLogger(ctx)
+	clients, err := client.GetClients(ctx, logger)
+	if err != nil {
+		return nil, "", "", err
+	}
+	defer clients.GRPCConn.Close()
```
jvallesm, line 85: Relying on the code using `client.GetClients` to remember about closing `GRPCConn` (which isn't used for anything else) is rather optimistic. Imagine some package just needs the `client.BlobStorageClient`, it's likely the dev won't even consider closing this.

pkg/data/file.go
chuang8511: Is it like what I proposed a few days ago?
image

donch1989: Another approach is we can make the BlobFetcher be a singleton.

jvallesm: Actually I saw no clients for `NewFileFromURL` (will it be used?), but I do see clients for other functions that use the `convertURLToBytes`, so the question remains valid.

Instead of passing the clients as function params, I see much more natural creating a `BlobFetcher` or `BinaryFetcher` (to align with current naming, i.e., `NewBinaryFromURL`) that acts as the receiver of the `ConvertURLToBytes` (I wrote it as exported because we might want to place this in another package, not sure about that).

```go
type BinaryFetcher struct {
	HTTPCLient     *resty.Client
	ArtifactClient artifactpb.ArtifactPrivateServiceClient // By having this injected, main.go is responsible of closing the connection.
	MinIOClient    miniox.MinioI
}

func (f *BinaryFetcher) FetchFromURL(url string) (b []byte, contentType string, filename string, err error) {
	if strings.HasPrefix(url, "data:") {
		return w.convertDataURIToBytes(url)
	}
	if matches := minioURLPattern.FindStringSubmatch(url); matches != nil {
		if len(matches) < 2 {
			return nil, "", "", fmt.Errorf("invalid blob storage url: %s", url)
		}
		return w.fetchFileFromBlobStorage(uuid.FromStringOrNil(matches[1]))
	}
	return w.fetchExternalURL(url)
}
```
Then, from Pipeline we can do the following (e.g. for the [audio](https://github.com/instill-ai/pipeline-backend/blob/main/pkg/service/pipeline.go) or the [binary](https://github.com/instill-ai/pipeline-backend/blob/main/pkg/service/pipeline.go) fetchers):
```go
			case "audio", "audio/*":
				url := defaultValueMap[k].(string)
				if v != nil {
					url = v.GetStringValue()
				}
				data, err := s.binaryFetcher.FetchFromURL(url)
				if err != nil {
					return fmt.Errorf("fetching binary data from URL: %w", err)
				}

				variable[k], err = data.CreateAudioData(data)
				if err != nil {
					return fmt.Errorf("creating audio data: %w", err)
				}
			case "file", "*/*":
				url := defaultValueMap[k].(string)
				if v != nil {
					url = v.GetStringValue()
				}
				data, err := s.binaryFetcher.FetchFromURL(url)
				if err != nil {
					return fmt.Errorf("fetching binary data from URL: %w", err)
				}

				variable[k], err = data.NewBinaryFromBytes(data) // And we can remove NewBinaryFromURL
				if err != nil {
					return fmt.Errorf("creating audio data: %w", err)
				}

```

donch1989: I think introducing a BinaryFetcher is a good idea. However, there are two concerns:

1. We might need to create a file from a URL within components, which would require injecting the BlobFetcher into each component implementations.
2. We also create files by URL in the data marshalling functions, meaning the BlobFetcher would need to be injected there as well.

If we can make BinaryFetcher a global singleton and initialize it in main.go, I guess it could simplify things.

pkg/worker/workflow.go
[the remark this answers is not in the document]
jvallesm: 🤔 Perhaps I'm misunderstanding what we are uploading. This is the trigger input, right? Basically the same that is uploaded already to MinIO to display the input of a pipeline run. If this is true, the trigger input should be placed (and consume) the requester's storage, not the owners. On Cloud, imagine a Free user (whose tier has a limitation on MinIO storage) that creates a very popular pipeline, the triggers would fill up their space eventually prevent other users from triggering the pipeline.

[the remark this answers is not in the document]
donch1989: I believe we are uploading the pipeline run’s output here, not its input. The input will be handled directly by the Console.

pkg/data/file.go
jvallesm: My concern with the singleton is that it hides these dependencies. Especially with the 1st point, this would prevent having components as standalone executors (using the package to execute the task or a CLi like `instill component execute openai --task=TEXT_GENERATION`) as sometimes we've discussed. The dangers of singletons are well documented so I'm against them using them lightly (I reckon there are cases where they make sense).

jvallesm: Well, to be a bit fair, it won't _prevent_ that use case but I imagine the dev experience being frustrating because you didn't know you needed to initialise that singleton.

pkg/worker/workflow.go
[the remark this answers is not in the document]
jvallesm: Correct, but it still applies. Regardless input / output (console / VDP), the data belongs to the requester right?

pkg/data/file.go
donch1989: Yeah, make senses

donch1989: Then I guess we probably need to make data.Marshal and data.Unmarhsal to be something like this
```golang
marshaler := data.NewMarshaler().SetBinaryFetcher(binaryFetcher)
value, _ := marshaler.marshal(data)
```

pkg/worker/workflow.go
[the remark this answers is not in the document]
donch1989: Indeed.

pkg/data/file.go
jvallesm: Looks good 👍 Also, we can always default to the HTTP fetcher, the artifact object conversion + minIO fetch is only an optimisation to save egress costs. That way, the use cases you mentioned would work without any extra plumbing.

pkg/worker/workflow.go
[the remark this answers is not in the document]
chuang8511: Like [this](https://github.com/instill-ai/pipeline-backend/pull/834/commits/8560348736f5fd61ee3a215da68b4c8e88e89d33). right?

I guess the code was wrong previously?
They should be requester namespace and UID. 
When the pipelines are triggered, the namespace from the API will be always requester. right?

```
				PipelineOwnerType:    ns.NsType,
				PipelineOwnerUID:     ns.NsUID,
```

pkg/data/client/client.go
chuang8511: I forgot to remove it because originally I want to use singleton.

pkg/service/main.go
chuang8511: Originally, I prepared it because I think service will need to upload and fetch the blob data in the future. 
But, now, I guess we won't need it after we change according to the discussion.


However, to be honest..., I am not sure what is the conclusion in the thread below. I will check it deeper.

pkg/data/file.go
chuang8511:
```
marshaler := data.NewMarshaler().SetBinaryFetcher(binaryFetcher)
value, _ := marshaler.marshal(data)
```

Not so sure what it means here.

Where do we init this marshaler?

pkg/service/main.go
chuang8511: I will remove it out first. And, in the future, when we need it, let's add it back later.

pkg/data/file.go
chuang8511: I exposed the Close func for the Clients.

chuang8511:
@jvallesm @donch1989 
I will deal with blob fetcher in ins-6872 when I confirm what you guys mean there. 
And, I have modified all other issues. Please take a look on them.

pkg/resource/resource.go
```diff
@@ -60,9 +60,9 @@ const (
 // doesn't matter whether the namespace belongs to a user or organization. This
 // refactor should be completed by August 2024.
 type Namespace struct {
```
chuang8511, line 62: Not sure if it will be better to make the system variables flatten rather than nested.

chuang8511: Because of this change, we also have to changed the pipeline-backend-cloud

pkg/data/file.go
jvallesm: Since this won't be tackled in this PR, I responded in the ticket with a more detailed proposal.

pkg/data/file.go
```diff
@@ -56,7 +66,55 @@ func convertURLToBytes(url string) (b []byte, contentType string, filename strin
...
+
+	objectRes, err := artifactClient.GetObject(ctx, &artifactpb.GetObjectRequest{
+		Uid: objectUID,
+	})
+
+	if err != nil {
+		return nil, "", "", err
+	}
+
+	// TODO: we have agreed on to add the bucket name in pb.Object
```
jvallesm, line 105: 👍 Thanks for documenting this

pkg/recipe/variable.go
```diff
@@ -14,14 +14,15 @@ import (
...
-	ExpiryRuleTag      string                 `json:"__EXPIRY_RULE_TAG"`
+	PipelineTriggerID  string    `json:"__PIPELINE_TRIGGER_ID"`
+	PipelineID         string    `json:"__PIPELINE_ID"`
+	PipelineUID        uuid.UUID `json:"__PIPELINE_UID"`
+	PipelineReleaseID  string    `json:"__PIPELINE_RELEASE_ID"`
+	PipelineReleaseUID uuid.UUID `json:"__PIPELINE_RELEASE_UID"`
+	ExpiryRuleTag      string    `json:"__EXPIRY_RULE_TAG"`
+
+	// PipelineNamespace is the namespace of the requester of the pipeline.
+	PipelineNamespace resource.Namespace `json:"__PIPELINE_NAMESPACE"`
```
jvallesm, line 25: I agree that if we need to know the UID and owner type of the requester a `resource.Namespace` is better than 2 fields. However, I see 2 problems:
- `PipelineNamespace` isn't self-documented, i.e., if this is used along the code I'll need to go to to the documentation to see whose namespace this is. By default I'd think this is the pipeline owner's namespace, as the resource path for a pipeline is `/namespaces/{namespaceID}/pipelines/{pipelineID}`. I'd use `PipelineRequester` as a name instead.
- This field is redundant with `PipelineRequesterUID`, which should be removed (or at least marked as deprecated for now and cleaned up in a subsequent PR).
```suggestion
	PipelineRequester resource.Namespace `json:"__PIPELINE_REQUESTER"`
```

pkg/repository/client.go
jvallesm, line 1: nit: sorry, as mentioned in the ticket I think the best location for this is `pkg/client` (which would be a rename of `pkg/external`). Can be addressed when this is refactored but I wanted to add a note on this.

pkg/resource/resource.go
jvallesm: If there are no blockers for the nested structure, I prefer it because it groups fields that are related (otherwise we duplicate `vars.PipelineRequesterType`, `vars.PipelineRequesterUID`, etc.).

Adding the JSON tags here adds some coupling but I'm fine with it. But perhaps we should follow the convention in `pkg/recipe` and have the tags in UPPER_SNAKE_CASE

jvallesm (commented):
Some comments, nothing serious. This is pretty much ready to go.

donch1989:
I'll merge this PR first. We can address the optimization and refactor issue later.

PR #835, opened 2024-11-12
feat(recipe): refactor run-on-event recipe structure

donch1989:
Because

- Previously, we used the following format to set up the run-on-event and run-on-schedule recipes. For certain application events, we'll use a polling method to capture the event, which overlaps with the run-on-schedule functionality.
    ```
    on:
      event:
        <event-id>:
          type: slack
          event: EVENT_NEW_MESSAGE
      schedule:
        <schedule-id>:
          cron:
    ```

This commit

- merges run-on-event and run-on-schedule and refactors recipe structure to use the following new format
    ```
    on:
      <event-id>:
        type: slack
        event: EVENT_NEW_MESSAGE
        config:
          channel: xxx
        setup: ${connection.xxx}
    
      <schedule-id>:
        type: schedule
        event: EVENT_SCHEDULE
        config:
          schedule: 
            cron: 
              - '30 2 * * *' 
    ```

### DIFF
--- a/pkg/datamodel/datamodel.go
+++ b/pkg/datamodel/datamodel.go
@@ -289,9 +289,10 @@ type Output struct {
 }
 
 type Event struct {
-	Type  string         `json:"type,omitempty" yaml:"type,omitempty"`
-	Event string         `json:"event,omitempty" yaml:"event,omitempty"`
-	Setup map[string]any `json:"setup,omitempty" yaml:"setup,omitempty"`
+	Type   string         `json:"type,omitempty" yaml:"type,omitempty"`
+	Event  string         `json:"event,omitempty" yaml:"event,omitempty"`
+	Config map[string]any `json:"config,omitempty" yaml:"config,omitempty"`
+	Setup  any            `json:"setup,omitempty" yaml:"setup,omitempty"`
 }
 
 type Schedule struct {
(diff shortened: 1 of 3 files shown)


pkg/datamodel/datamodel.go
```diff
@@ -288,15 +288,11 @@ type Output struct {
...
-	Schedule map[string]*Schedule `json:"schedule,omitempty" yaml:"schedule,omitempty"`
-}
-
 type Event struct {
-	Type  string         `json:"type,omitempty" yaml:"type,omitempty"`
-	Event string         `json:"event,omitempty" yaml:"event,omitempty"`
-	Setup map[string]any `json:"setup,omitempty" yaml:"setup,omitempty"`
+	Type   string         `json:"type,omitempty" yaml:"type,omitempty"`
+	Event  string         `json:"event,omitempty" yaml:"event,omitempty"`
+	Config map[string]any `json:"config,omitempty" yaml:"config,omitempty"`
```
donch1989, line 294: Hi @jvallesm,

I’d like to discuss an idea with you.

I’m currently refactoring the run-on-event recipe, which requires both an event-specific configuration and a connection setup. Here’s an example:
```
on:
  <event-id>:
    type: slack
    event: EVENT_NEW_MESSAGE
    config:
      channel: xxx
    setup: ${connection.my-slack}
```
Having both config and setup feels a bit confusing. Would it be clearer if we renamed setup to connection?

```
on:
  <event-id>:
    type: slack
    event: EVENT_NEW_MESSAGE
    config:
      channel: xxx
    connection: ${connection.my-slack}
```
If so, we’ll also need to update the `setup` field to `connection` in our component tasks for consistency. What do you think?

PR #845, opened 2024-11-18
chore(integration): whitelist certain user emails for certain integrations

jvallesm:
Because

- Certain OAuth integrations like Google Drive pass a review process. We need to test the scopes internally before making the app public.

This commit

- Adds a way to hide the OAuth integration for certain components and emails.


jvallesm:
# QA ✅ 

On CE, we only have the `admin` user, so I needed to update its email in the database. Then, `pipeline-backend` was launched as:

```sh
export CFG_COMPONENT_INTERNALUSEREMAILS=jvm@gmail.com,a09@gmail.com e52@gmail.com
export CFG_COMPONENT_COMPONENTSWITHINTERNALUSERS=google-drive,github
go run ./cmd/maingg
```

The `oAuthConfig` in `8080/v1beta/integrations/<id>` was toggled by the email in the DB and component ID.

donch1989 (dismissed):

PR #847, opened 2024-11-18
fix(service): skip empty component definition in API response

donch1989:
Because

 - Previously, the instill-app component was excluded, resulting in an empty item in the API response, which caused the Console to fail.
image

This commit

- skips empty component definition in API response


donch1989:
@chuang8511 
I fix this problem that causing Console failing.

jvallesm (dismissed):

chuang8511:
I put this logic in service layer because I felt it was not a repo logic.
But, in long run, we will move the cloud-only component into another repository.
So, I guess it will be fine with the current way.

PR #866, opened 2024-11-20
fix(github): fix time format bug

chuang8511:
Because

- in ymal parser, we only provide date only string format

This commit

- convert GitHub DateTime input to DateOnly

Note
- We will need to find a solution about how to provide the Date and DateTime at the same time. Now, the current YAML parser cannot read the original users' input.

### DIFF
--- a/pkg/component/application/github/v0/README.mdx
+++ b/pkg/component/application/github/v0/README.mdx
@@ -296,7 +296,7 @@ Get the review comments in a pull request. The comments can be on a specific lin
 | PR Number | `pr-number` | integer | Number of the PR. Default is `0`, which retrieves all comments on all PRs in the repository. |
 | Sort | `sort` | string | Sort the comments by created, updated. Default is created. |
 | Direction | `direction` | string | Direction of the sort, including asc or desc. Default is desc. |
-| Since | `since` | string | Only comments updated at or after this time are returned. Default is 2021-01-01T00:00:00Z. |
+| Since | `since` | string | Date (in `YYYY-MM-DD` format) from which comments will start to be fetched. The date will be in the UTC timezone. |
 | Page | `page` | integer | Page number of the results to fetch. Default is 1. |
 | Per Page | `per-page` | integer | Number of results to fetch per page. Default is 30. |
 </div>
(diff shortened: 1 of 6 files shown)


jvallesm (approved):
Thanks for fixing this. Accepting only one format isn't ideal, and to me a full datetime makes more sense than a plain date, but if the UX is better with the latter, according to product, let's go ahead with it. My concern is that dates don't have a timezone so it will be interpreted as UTC, which might not be clear to the user (even if it's documented, thanks for taking care of that). I do get the point, though, that a timestamp or a RFC3339 datetime are more cumbersome to write than a date from the user perspective.

> We will need to find a solution about how to provide the Date and DateTime at the same time. Now, the current YAML parser cannot read the original users' input.

We could have a list of accepted formats (it may contain only datetime and RFC3339, perhaps also datetime with timezone as a custom format).

donch1989:
> Thanks for fixing this. Accepting only one format isn't ideal, and to me a full datetime makes more sense than a plain date, but if the UX is better with the latter, according to product, let's go ahead with it. My concern is that dates don't have a timezone so it will be interpreted as UTC, which might not be clear to the user (even if it's documented, thanks for taking care of that). I do get the point, though, that a timestamp or a RFC3339 datetime are more cumbersome to write than a date from the user perspective.
> 
> > We will need to find a solution about how to provide the Date and DateTime at the same time. Now, the current YAML parser cannot read the original users' input.
> 
> We could have a list of accepted formats (it may contain only datetime and RFC3339, perhaps also datetime with timezone as a custom format).

I agree that we should make it more tolerant. I’m thinking we can introduce a new Instill format for datetime in the future. We can leverage the auto-conversion capabilities provided by our Instill format framework for this.

jvallesm:
> We could have a list of accepted formats (it may contain only datetime and RFC3339, perhaps also datetime with timezone as a custom format).

I left my sentence unfinished, sorry. To clarify it, what I meant by this is ranging over a list of formats and failing only when parsing failed for all of them.

chuang8511:
> Accepting only one format isn't ideal, and to me a full datetime makes more sense than a plain date, but if the UX is better with the latter, according to product, let's go ahead with it.

Yeah, I was thinking that only accepting one format isn't good. But, I just thought about if the users really need DateTime in this specific GitHub use cases. So, as a hotfix bug, I chose the current solution.

In long run, I also agree that we should accept more different types.

PR #872, opened 2024-11-24
refactor(component,github): adopt new codebase structure

pinglin:
Because

- we'd like to have each component codebase follow consistent structure and style

This commit

- refactor file and codebase structure
- make tasks in alphabetical order in `tasks.json`, `main.go`, and `io.go`
- use ConcurrentExecutor for component execution


pkg/component/base/execution.go
```diff
@@ -29,28 +29,36 @@ type IExecution interface {
...
 }
 
+// Job is the job for the component.
 type Job struct {
 	Input  InputReader
 	Output OutputWriter
 	Error  ErrorHandler
 }
 
+// InputReader is an interface for reading input data from a job.
```
jvallesm, line 39: Thanks for this, my IDE (also) complains about uncommented exported types and functions. I've tried to remove such warnings as I modify packages as you're doing with this PR 👍 👍

pkg/component/base/execution.go
```diff
@@ -333,7 +367,7 @@ func Validate(data *structpb.Struct, jsonSchema string, target string) error {
 // initialization.
 const SecretKeyword = "__INSTILL_SECRET"
 
-// NewUnresolvedCredential returns an end-user error signaling that the
+// NewUnresolvedCredential returns an end-user error signalling that the
```
jvallesm, line 370: I'd used the US spelling intentionally as it seems to be what we use in the documentation.

pkg/component/base/execution.go
```diff
@@ -121,13 +139,16 @@ func (e *ComponentExecution) getInputSchemaJSON(task string) (map[string]interfa
 	inputMap := taskSpecMap["properties"].(map[string]interface{})
 	return inputMap, nil
 }
-func (e *ComponentExecution) FillInDefaultValues(input *structpb.Struct) (*structpb.Struct, error) {
+
+// FillInDefaultValues fills in the default values for the input based on the
```
jvallesm, line 143: 👀 The comment is incomplete

pkg/component/application/jira/v0/component_test.go
```diff
@@ -35,17 +35,29 @@ func TestComponent_ListBoardsTask(t *testing.T) {
...
 				StartAt:    0,
 			},
 			wantResp: ListBoardsOutput{
-				Total:      1,
+				Total:      3,
 				StartAt:    0,
 				MaxResults: 10,
 				IsLast:     true,
 				Boards: []Board{
+					{
```
jvallesm, line 49: @Ping-Lin Chang this means that your implementation of `FillInDefaultValues` isn't filling the default value for the `board-type` parameter (`"simple"`). Indeed there are three boards but the result should return only one because the [task by default filters the simple boards](https://github.com/instill-ai/pipeline-backend/blob/main/pkg/component/application/jira/v0/config/tasks.json#L227).

pinglin: Good point. I will refactor `FillInDefaultValues` completely and ensure this test can pass originally.

pkg/component/base/execution.go
```diff
@@ -162,94 +184,173 @@ func optionMatch(valueMap *structpb.Struct, reference map[string]interface{}, ch
...
+		}
+	}
+
+	inputStruct.GetFields()[key] = structpb.NewListValue(tempArray)
+	return nil
+}
+
+// Main function
+func (e *ComponentExecution) fillInDefaultValuesWithReference(input any, reference map[string]interface{}) (any, error) {
+	inputStruct, err := convertToStructPb(input)
```
jvallesm, line 319: nit: if this is only going to take and return a `structpb.Struct`, does it make sense to have `any` in the signature instead of that type?

jvallesm (dismissed):
I added some minor comments, none of them are blockers to move forward 👍

pkg/component/base/execution.go
pinglin: This is a bit tricky because we need to pass input `any` from [here](https://github.com/instill-ai/pipeline-backend/pull/872/files#diff-af6972698e144d60b8b3202a679de0954cca17bd5d92a3d3bc1f7a5debd93c21R138). If we convert input from `any` to `*structpb.Struct` there, we will need to unmarshal [this](https://github.com/instill-ai/pipeline-backend/pull/872/files#diff-27f7ebfb46d49db462803ce492d42f1af3cfebbc3efc7956a302d55a7ecff01dR346-R348) in the same `Execute` function.

I think we'd better refactor how default values are injected to the input variable given the new I/O interface. IMHO, the current way is not elegant. cc'd @donch1989

jvallesm (dismissed):

pkg/component/base/execution.go
pinglin: > One thing I noticed is that every call to FillInDefaultValues will ignore the return value and only check the error. If that's the (only) intendeded use, perhaps we should simplify the signature. But we can do it when we refactor this part.

Good point. That was for the old implementation using with the internal recursive private function `fillInDefaultValuesWithReference`. I will take it out and rebase the latest main before merging.

donch1989: The fillInDefaultValuesWithReference indeed not works well under the current structure. I'm thinking we can utilize the go tag with our unmarhal function to handle this.
For example
```golang
type Output struct{
  str1 *string `instill:"str1,default=VAL"`
}
```

PR #875, opened 2024-11-26
test(integration): fix test cases for integrations

donch1989:
Because

- The original test case for integration failed due to:
  1. Removal of the connector-definition endpoint
  2. Changes in OAuth scope

This commit:

- Fixes the test cases for integrations

Note:

- Investigate why the test did not raise an error in CI.

### DIFF
--- a/integration-test/pipeline/rest-integration.js
+++ b/integration-test/pipeline/rest-integration.js
@@ -27,8 +27,16 @@ export function CheckIntegrations() {
     });
 
     var id = "github";
-    var cdef = http.request("GET", `${pipelinePublicHost}/v1beta/connector-definitions/${id}`, null, null).
-      json().connectorDefinition;
+    var cdefs = http.request("GET", `${pipelinePublicHost}/v1beta/component-definitions?pageSize=100`, null, null).
+      json().componentDefinitions;
+
+    var cdef = null;
+    for (var i = 0; i < cdefs.length; i++) {
+      if (cdefs[i].id === id) {
+        cdef = cdefs[i];
+        break;
+      }
+    }
 
     var integration = {
       uid: cdef.uid,
@@ -45,7 +53,7 @@ export function CheckIntegrations() {
     var oAuthConfig = {
       authUrl: "https://github.com/login/oauth/authorize",
       accessUrl: "https://github.com/login/oauth/access_token",
-      scopes: ["repo", "write:repo_hook"],
+      scopes: ["repo", "admin:repo_hook"],
     };
 
     // Basic view
--- a/integration-test/pipeline/rest-integration.js
+++ b/integration-test/pipeline/rest-integration.js
@@ -27,7 +27,7 @@ export function CheckIntegrations() {
     });
 
     var id = "github";
-    var cdefs = http.request("GET", `${pipelinePublicHost}/v1beta/component-definitions?pageSize=100`, null, null).
+    var cdefs = http.request("GET", `${pipelinePublicHost}/v1beta/component-definitions?filter=qTitle="GitHub"`, null, null).
       json().componentDefinitions;
 
     var cdef = null;


integration-test/pipeline/rest-integration.js
```diff
@@ -27,8 +27,16 @@ export function CheckIntegrations() {
     });
 
     var id = "github";
-    var cdef = http.request("GET", `${pipelinePublicHost}/v1beta/connector-definitions/${id}`, null, null).
-      json().connectorDefinition;
+    var cdefs = http.request("GET", `${pipelinePublicHost}/v1beta/component-definitions?pageSize=100`, null, null).
```
jvallesm, line 30: 👍 I absolutely overlooked this, it should have been `component-definitions` since the beginning.

Instead of looping over the results, you can use the `filter` query param:

```suggestion
    var cdefs = http.request("GET", `${pipelinePublicHost}/v1beta/component-definitions?filter=qTitle="GitHub"`, null, null).
```

donch1989: Yes, that's better. Updated.

jvallesm (approved):

donch1989:
> > Investigate why the test did not raise an error in CI.
> 
> Just panicking won't return an error for the command (which is odd as behaviour), and we're failing outside of a check. There are 2 approaches here:
> 
> ### Checking the response before using it
> This is more specific to our problem and good practice in general. We're fetching the information assuming the HTTP request will succeed, which we shouldn't do.
> 
> ```js
>     var id = "github";
>     var cdResp = http.request("GET", `${pipelinePublicHost}/v1beta/connector-definitions/${id}`, null, null);
>     check(cdResp, {
>       "GET /v1beta/connector-definitions/github response status is 200": (r) => r.status === 200,
>     });
>     var cdef = cdResp.json().connectorDefinition;
> ```
> 
> ### Using `describe()`
> As mentioned in [Error Handling](https://grafana.com/docs/k6/latest/javascript-api/jslib/k6chaijs/error-handling/#handling-exceptions), we can use `describe` instead of `group` (but it makes use another import). Considering we can introduce a panic error anywhere, perhaps this is the most robust approach.
> 
> ```js
> import http from "k6/http";
> import { check, group } from "k6";
> import { randomString } from "https://jslib.k6.io/k6-utils/1.1.0/index.js";
> import { describe, expect } from 'https://jslib.k6.io/k6chaijs/4.3.4.3/index.js';
> 
> // ...
> 
>   describe("Integration API: Get integration", () => {
>     // Inexistent component
>     check(http.request("GET", `${pipelinePublicHost}/v1beta/integrations/restapio`, null, null), {
>       "GET /v1beta/integrations/restapio response status is 404": (r) => r.status === 404,
>       "GET /v1beta/integrations/restapio response contains end-user message": (r) => r.json().message === "Integration does not exist.",
>     });
> 
>     // Component without setup
>     check(http.request("GET", `${pipelinePublicHost}/v1beta/integrations/document`, null, null), {
>       "GET /v1beta/integrations/document response status is 404": (r) => r.status === 404,
>       "GET /v1beta/integrations/document response contains end-user message": (r) => r.json().message === "Integration does not exist.",
>     });
> 
>     var id = "github";
>     var cdef = http.request("GET", `${pipelinePublicHost}/v1beta/connector-definitions/${id}`, null, null).
>       json().connectorDefinition;
> ```
> 
> The output in this case is:
> 
> ```
>      █ Integration API: Get integration
> 
>        ✓ GET /v1beta/integrations/restapio response status is 404
>        ✓ GET /v1beta/integrations/restapio response contains end-user message
>        ✓ GET /v1beta/integrations/document response status is 404
>        ✓ GET /v1beta/integrations/document response contains end-user message
>        ✗ Exception raised "GoError: cannot parse json due to an error at line 1, character 6 , error: invalid character 'p' after top-level value"
>         ↳  0% — ✓ 0 / ✗ 1
> 
>      █ Integration API: List integrations
> 
>        ✓ GET /v1beta/integrations response status is 200
>        ✓ GET /v1beta/integrations response totalSize > 0
>        ✓ GET /v1beta/integrations has default page size
>        ✓ GET /v1beta/integrations?pageSize=2&pageToken=eyJzY29yZSI6MCwidWlkIjoiY2QyMjBkMmQtM2QxOS00NjhlLThiOTUtMzdkZDZhNTdjMTVmIn0= response status is 200
>        ✓ GET /v1beta/integrations?pageSize=2&pageToken=eyJzY29yZSI6MCwidWlkIjoiY2QyMjBkMmQtM2QxOS00NjhlLThiOTUtMzdkZDZhNTdjMTVmIn0= has page size 2"
>        ✓ GET /v1beta/integrations?pageSize=2&pageToken=eyJzY29yZSI6MCwidWlkIjoiY2QyMjBkMmQtM2QxOS00NjhlLThiOTUtMzdkZDZhNTdjMTVmIn0= has different elements than page 1"
>        ✓ GET /v1beta/integrations?filter=qIntegration="que" response status is 200
>        ✓ GET /v1beta/integrations?filter=qIntegration="que" response totalSize > 0
> ...
> ..
> .
> ERRO[0006] thresholds on metrics 'checks' have been crossed
> ```

Thanks for sharing. We need to be cautious about this. It’s too terrible that the test failed but didn’t raise an error.

PR #879, opened 2024-11-27
feat(smartlead): implement smartlead

chuang8511:
Because

- we want to build agent

This commit

- implement smartlead
- fix compogen bug by add enum to input


chuang8511:
If there is anything we need to modify, I will modify it in another PR. So, I merge it first

PR #882, opened 2024-11-28
feat(component,openai): support streaming for o1-mini and o1-preview models

donch1989:
Because

- OpenAI add support for streaming on the o1 models

This commit

-  supports streaming for o1-mini and o1-preview models


chuang8511:
Ah, we will need to do v1 as well for universal AI component.

donch1989:
> Ah, we will need to do v1 as well for universal AI component.

Right, please help update it when you have time. Thanks.

PR #883, opened 2024-11-28
feat(trigger): accept connection references in the pipeline trigger data

jvallesm:
Because

- When a pipeline recipe references a connection, that connection ID will
  be fetched from the requester's namespace.
- This forces requesters to create connections in order to trigger
  pipelines from other namesapces.

This commit

- Adds a `connectionReferences` field in the pipeline trigger data, which
  allows clients to override the references in the recipe with per-batch
  references.
- Adds a private endpoint to fetch connections by UID, needed in the
  1-minute to impact project.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -38,7 +38,7 @@ require (
 	github.com/h2non/filetype v1.1.3
 	github.com/iancoleman/strcase v0.3.0
 	github.com/influxdata/influxdb-client-go/v2 v2.12.3
-	github.com/instill-ai/protogen-go v0.3.3-alpha.0.20241127083047-e2cb3b0062e6
+	github.com/instill-ai/protogen-go v0.3.3-alpha.0.20241128113411-7d041cf4df22
 	github.com/instill-ai/usage-client v0.2.4-alpha.0.20240123081026-6c78d9a5197a
 	github.com/instill-ai/x v0.5.0-alpha.0.20241119141833-e4a78ca87792
 	github.com/itchyny/gojq v0.12.14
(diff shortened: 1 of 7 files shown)


jvallesm:
## QA ✅ 

With the following recipe:
```yaml
variable:
  topic:
    format: string
component:
  openai-0:
    type: openai
    input:
      model: gpt-4o
      prompt: Tell me a joke about ${variable.topic}
      system-message: You are a helpful assistant.
      temperature: 1
      n: 1
      max-tokens:
      response-format:
        type: text
      top-p: 1
    task: TASK_TEXT_GENERATION
    setup: ${connection.oai}

output:
  out-topic:
    value: ${openai-0.output.texts}
  out-other:
    value: ${openai-1.output.texts}
```

```sh
curl -X POST 'http://localhost:8080/v1beta/users/admin/pipelines/openai-joke/trigger' \
  --header "Content-Type: application/json" \
  --header "Authorization: Bearer $INSTILL_API_TOKEN" \
  --data '{
  "data": [
    {"variable": {"topic": "cats"}, "connectionReferences": {"oai": "my-openai"}},
    {"variable": {"topic": "musicians"}, "connectionReferences": {"oai": "my-openai"}}
  ]
}' | jq
```

- Without `connectionReferences` and no `oai` connection in the namespace, "connection doesn't exist" error ✅ 
- With `connectionReferences` we can override the `oai` connection, whether it exists or not ✅ 
- With several components referencing the same connection, the connection is fetched from the DB only once ✅ 
- With several batches referencing the same connection, the connection is fetched from the DB only once ✅

donch1989 (approved):
LGTN

PR #898, opened 2024-12-02
chore(perplexity): rename perplexity

chuang8511:
Because

- the name is wrong

This commit

- rename perplexity


chuang8511:
### Notice!
This should be deployed with terraform-instill-cloud-gcp-services pr-302.

PR #899, opened 2024-12-02
feat(vdp): skip google drive for cloud version

chuang8511:
Because

- we want to hide google drive in cloud before we get drive.readonly scope from Google

This commit

- hide google drive in cloud version

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -17,6 +17,13 @@ import (
 	miniox "github.com/instill-ai/x/minio"
 )
 
+const (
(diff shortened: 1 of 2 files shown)


jvallesm (approved):
👍 I'm fine with this quick way to achieve the goal of hiding Google Drive in cloud.

In the future I'd have separate `store.Init` functions for CE and Cloud in order to decouple both distributions.

donch1989 (approved):

PR #904, opened 2024-12-02
feat(vdp): upload raw inputs for run log

chuang8511:
Because

- now, the run log inputs are raw data but not blob storage
- we want the inputs are download urls rather than dataURI

This commit

- move upload pipeline run inputs from worker to service
  - remove the worker upload inputs function
  - add the service upload inputs function


chuang8511:
A point to be discussed.
image

pkg/service/pipeline.go
```diff
@@ -1064,36 +1103,44 @@ func (s *service) preTriggerPipeline(ctx context.Context, ns resource.Namespace,
 					if err != nil {
 						return err
 					}
+					uploadingPipelineData[idx][k] = v.GetStringValue()
```
donch1989, line 1106: @chuang8511 
The URL might still be a dataURI (base64). If the user continues using dataURI or base64 in the input, we should save it to our blob storage and replace it with the corresponding blob URL. Could you check if this change can be implemented quickly? I guess we can modify lines L868–L907 to handle this.

chuang8511: > The URL might still be a dataURI (base64).

Do you mean when the users call API to trigger pipeline but not from Console?

donch1989: Yes, that’s one of the possibilities.

chuang8511: fixed in [this commit](https://github.com/instill-ai/pipeline-backend/pull/904/commits/8b5decb2e659173b0e044753eee49f17c6751670).

pkg/service/pipelinerun.go
```diff
@@ -380,3 +382,47 @@ func (s *service) ListPipelineRunsByRequester(ctx context.Context, req *pb.ListP
...
+
+type uploadPipelineRunInputsToMinioParam struct {
+	PipelineTriggerID string
+	ExpiryRuleTag     string
+	PipelineData      []map[string]any
+}
+
+func (s *service) uploadPipelineRunInputsToMinio(ctx context.Context, param uploadPipelineRunInputsToMinioParam) error {
+	logger, _ := logger.GetZapLogger(ctx)
+	minioClient := s.minioClient
```
jvallesm, line 394: nit: not something you introduced but I don't love having the `MinIO` client as a direct dependency in the domain layer.

pkg/service/blob_storage.go
jvallesm, line 1: nit: I have no strong opinion for underscored file names, but `pipelinerun.go` and `blob_storage.go` follow different conventions. The latter seems to be the most extended.

pkg/service/main.go
```diff
@@ -98,37 +99,41 @@ type TriggerResult struct {
...
+	aclClient                    acl.ACLClientInterface
+	converter                    Converter
+	minioClient                  miniox.MinioI
+	memory                       memory.MemoryStore
+	log                          *zap.Logger
+	workerUID                    uuid.UUID
+	retentionHandler             MetadataRetentionHandler
+	binaryFetcher                external.BinaryFetcher
+	artifactPublicServiceClient  artifactpb.ArtifactPublicServiceClient
+	artifactPrivateServiceClient artifactpb.ArtifactPrivateServiceClient
```
jvallesm, line 117: ❓ Is this one used?

pkg/service/blob_storage.go
```diff
@@ -0,0 +1,114 @@
...
+
+	ctx = metadata.NewOutgoingContext(ctx, utils.GetRequestMetadata(vars))
+
+	timestamp := time.Now().Format(time.RFC3339)
+	objectName := fmt.Sprintf("%s-%s%s", requesterUID.String(), timestamp, getFileExtension(mimeType))
+
+	resp, err := artifactClient.GetObjectUploadURL(ctx, &artifactpb.GetObjectUploadURLRequest{
+		NamespaceId:      ns.NsID,
+		ObjectName:       objectName,
+		ObjectExpireDays: 0,
```
jvallesm, line 44: If I understand correctly, we only use this to upload input data (e.g. an image) before continuing with the pipeline pre-trigger execution. Later, we upload the hole pipeline run input (all the input values, whether referenced through a URL or not) to Minio again. Why does this has no expiration and the pipeline run input doesn't?

pkg/service/blob_storage.go
```diff
@@ -0,0 +1,114 @@
...
+	if err != nil {
+		return "", fmt.Errorf("generate system variables: %w", err)
+	}
+
+	ctx = metadata.NewOutgoingContext(ctx, utils.GetRequestMetadata(vars))
+
+	timestamp := time.Now().Format(time.RFC3339)
+	objectName := fmt.Sprintf("%s-%s%s", requesterUID.String(), timestamp, getFileExtension(mimeType))
+
+	resp, err := artifactClient.GetObjectUploadURL(ctx, &artifactpb.GetObjectUploadURLRequest{
```
jvallesm, line 41: Can you refresh my memory about when (and why) we go through `artifact-backend` and when we use the MinIO client directly? 🙏

pkg/service/pipeline.go
```diff
@@ -866,38 +864,31 @@ func (s *service) preTriggerPipeline(ctx context.Context, ns resource.Namespace,
...
 						continue
 					}
-					if !strings.HasPrefix(s, "data:") {
-						b, err := base64.StdEncoding.DecodeString(s)
-						if err != nil {
-							return fmt.Errorf("can not decode file %s, %s", formatMap[k], s)
-						}
-						mimeType := strings.Split(mimetype.Detect(b).String(), ";")[0]
-						vars.Fields[k] = structpb.NewStringValue(fmt.Sprintf("data:%s;base64,%s", mimeType, s))
+					downloadURL, err := s.uploadBlobAndGetDownloadURL(ctx, ns, str)
```
jvallesm, line 874: ❓ Is every string interpreted as a data blob (and therefore uploaded)? E.g. in a text generation chat, will the chat input be uploaded in this fashion?

jvallesm (commented):
I tried to follow the changes as best as I could. You detailed the proposed flow in the ticket, could you use that to document what `preTriggerActivity` does? A comment in the method description and some comments along the function would be very helpful 🙏.

pkg/service/pipelinerun.go
chuang8511: I wonder why MinIO isn't in repository layer.
I think we should move MinIO client to repository layer. 
Is it same as what you think?

If yes, I will register a ticket for it.

pkg/service/main.go
chuang8511: Yes, it is used in blobstorage. In the future, I think it will be better to move it to repository layer if possible.

pkg/service/blob_storage.go
chuang8511: Oh, mainly, I copy my previous code from worker to here.
I will try to follow the way `uploadPipelineRunInputsToMinio` did.

If so, we may need to remove [artifact client from service layer](https://github.com/instill-ai/pipeline-backend/pull/904#discussion_r1869358566) as well.

pkg/service/blob_storage.go
chuang8511: 1. `NewMinioClientAndInitBucket` introduced by Jeremy for pipeline run log.
2. `NewMinioClient` introduced by me for pipeline output data which are blob data.

Now, there are 2 ways to upload data to blob storage.
1. Use `artifact-backend` to get the upload URL and `Put` the data with the upload URL.
2. Use minio client to upload data directly.

The reason why I did not follow the previous design is mainly that we need bucket name in the params when we `GetFile`. 
But in previous design, it assumes all data will be in the same bucket. So, I introduced `NewMinioClient`.

pkg/service/pipeline.go
chuang8511: No, only when the instill format meet this condition.
`formatMap[k] != "string" && formatMap[k] != "number" && formatMap[k] != "boolean" && formatMap[k] != "json" {`

I will use another function to make it more readable.

e.g. to follow the [term here](https://www.instill.tech/docs/vdp/instill-format#unstructured-data-types)
```
func isUnstructuredData(format string) {
   ...
}
```

pkg/service/blob_storage.go
chuang8511: Now, I realise the previous design.

In previous design, we upload and read data by the `minio client` initialised by `NewMinioClientAndInitBucket`.
This design mainly focused on uploading data and reading data internally.
Specifically, it means the input/output are only known by pipeline-backend itself but not the external client like user or console.
e.g. `pipeline-runs/input/d3e72143-0140-4224-ac89-9818a622df29.json`

Under the current design, we upload and read data by the `minio client` initialised by `NewMinioClient`.
This design mainly focused on uploading data and reading data for external usage.
e.g. `http://localhost:8080/v1alpha/namespaces/admin/blob-urls/0e39466e-515d-4449-af5b-e78ca6b994f8`

So, for the Pipeline input data, we upload those by `uploadPipelineRunInputsToMinio`, which is from the previous design. Those data should be read internally. So, we don't have to expose the download URL

But, for the user input data, it should be read by external users / console. So, we have to expose the download URL.

As for the retention issue, it should be addressed in another ticket.
Improvement: add retention when upload blob data in VDP (ins-6857).

Further description:

- Pipeline input data means the JSON object contains the data and metadata.
- User input data means the data users upload.

chuang8511: So, in summary, what we should is
- add the TODO in `ObjectExpireDays: 0,`, both `service` and `worker` pkg.
- write clearer document for this reason.

pkg/service/pipelinerun.go
jvallesm: Exactly, that was my thought ➕

pkg/service/pipeline.go
```diff
@@ -833,6 +831,17 @@ func (s *service) UpdateNamespacePipelineIDByID(ctx context.Context, ns resource
...
+//  1. Upload pipeline input data to minio if the data is blob data.
+//  2. New workflow memory.
+//     2-1. Set the default values for the variables for memory data and uploading pipeline data.
+//     2-2. Set the data with data.Value for the memory data, which will be used for pipeline running.
+//     2-3. Upload "uploading pipeline data" to minio for pipeline run logger.
+//  3. Map the settings in recipe to the format in workflow memory.
+//  4. Enable the streaming mode when the header contains "text/event-stream"
+//
+// We upload User Input Data by `uploadBlobAndGetDownloadURL`, which exposes the public URL because it will be used by `console` & external users.
+// We upload Pipeline Input Data by `uploadPipelineRunInputsToMinio`, which does not expose the public URL. The URL will be used by pipeline run logger.
```
jvallesm, line 844: Super clear! 🙇

pkg/service/blob_storage.go
jvallesm: > We upload Pipeline Input Data by `uploadPipelineRunInputsToMinio`, which does not expose the public URL. The URL will be used by pipeline run logger.

You mean that the response for the pipeline run list will return the data blob instead of the URL, right? And therefore it will need to fetch that info from MinIO while building the response.

pkg/service/pipeline.go
```diff
@@ -941,6 +948,7 @@ func (s *service) preTriggerPipeline(ctx context.Context, ns resource.Namespace,
 				// is missing and should be handled as such by components.
 				if d, ok := defaultValueMap[k]; !ok || d == nil {
 					variable[k] = data.NewNull()
+					uploadingPipelineData[idx][k] = nil
```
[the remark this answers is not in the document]
chuang8511: When the format is unstructured data, the data will be different.

e.g. [format image](https://github.com/instill-ai/pipeline-backend/blob/3e25aab56dce2745dfecbe05d9448bcb9dff7299/pkg/service/pipeline.go#L1093)
- variable -> will be blob data
- uploadingPipelineData[idx] -> will be the public URL that the console / users can access to.



Originally, I plan to write this part of code by
- Loop to set `uploadingPipelineData`.
- Upload data
- Loop to set `variable`

But, I think the current way is easier, so I did not choose my original plan.

pkg/service/blob_storage.go
jvallesm: 🤔 I think this is a leftover. In protobufs, `inputsReference` and `outputsReference`fields were [removed](https://github.com/instill-ai/protobufs/blob/main/vdp/pipeline/v1beta/pipeline.proto#L2067) from the `PipelineRun` message. I think their presence in `ComponentRun` is a leftover. I checked console and, though they're defined in the `PipelineRun` type (not in the component one), they aren't used.

pkg/service/pipeline.go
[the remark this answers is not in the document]
jvallesm: Yep my suggestion was only if the data was the same. The function is quite long so I missed these small details 😅 Thanks for the clarification, in that case the code makes sense as it is (we can work on it later to make it more readable and compact).

jvallesm (approved):

pkg/service/pipeline.go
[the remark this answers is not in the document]
chuang8511: recorded in ins-7071

PR #911, opened 2024-12-03
feat(component,instillapp): remove instill-app component

donch1989:
Because

- we retire instill-app component

This commit

- removes instill-app component


donch1989:
> Is there any document or ticket describing this decision? Sorry, I did not catch up this.

We discussed this in the VDP meeting yesterday. Due to the significant refactor planned for the app/agent-backend, the current implementation will not work.

PR #921, opened 2024-12-04
fix(repository): fix edition filter

donch1989:
Because

- we need to filter the integration table based on the edition.

This commit

- fixes the edition filter.
- removes the filter for instill-app component as it has been already removed.

### DIFF
--- a/pkg/repository/repository.go
+++ b/pkg/repository/repository.go
@@ -767,12 +767,8 @@ func (r *repository) ListComponentDefinitionUIDs(_ context.Context, p ListCompon
 		Where(where, whereArgs...).
 		Where("is_visible IS TRUE")
 
-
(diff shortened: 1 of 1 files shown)


pkg/repository/repository.go
```diff
@@ -878,6 +874,11 @@ func (r *repository) ListIntegrations(ctx context.Context, p ListIntegrationsPar
 		Where(where, whereArgs...).
 		Where("is_visible IS TRUE AND has_integration IS TRUE")
 
+	if config.Config.Server.Edition == config.EditionCloudStaging || config.Config.Server.Edition == config.EditionCloudProd {
```
chuang8511, line 877: Thanks. I missed the integration.

PR #923, opened 2024-12-04
feat(vdp): upload component input output data

chuang8511:
Because

- we want to integrate blob storage

This commit

- upload component run input/output data to blob storage


pkg/utils/blobstorage.go
```diff
@@ -0,0 +1,272 @@
...
+	for i, dataStruct := range params.DataStructs {
+		updatedDataStruct, err := uploadBlobDataAndReplaceWithURL(ctx, uploadBlobDataAndReplaceWithURLParams{
+			namespaceID:    params.NamespaceID,
+			requesterUID:   params.RequesterUID,
+			dataStruct:     dataStruct,
+			logger:         params.Logger,
+			artifactClient: params.ArtifactClient,
+		})
+		if err != nil {
+			// Note: we don't want to fail the whole process if one of the data structs fails to upload.
```
jvallesm, line 55: nit: you could use the `logger` you're passing around to signal that something failed but the execution continued.

pkg/utils/blobstorage.go
```diff
@@ -0,0 +1,272 @@
...
+
+type uploadBlobDataAndReplaceWithURLParams struct {
+	namespaceID    string
+	requesterUID   uuid.UUID
+	dataStruct     *structpb.Struct
+	logger         *zap.Logger
+	artifactClient *artifactpb.ArtifactPublicServiceClient
+}
+
+func uploadBlobDataAndReplaceWithURL(ctx context.Context, params uploadBlobDataAndReplaceWithURLParams) (*structpb.Struct, error) {
```
jvallesm, line 72: ❓ This method never returns an error, right?

pkg/worker/workflow.go
```diff
@@ -402,6 +398,13 @@ func (w *worker) TriggerPipelineWorkflow(ctx workflow.Context, param *TriggerPip
 			}
 			componentRunFutures = append(componentRunFutures, workflow.ExecuteActivity(minioCtx, w.UploadComponentOutputsActivity, futureArgs[idx]))
 		}
+
+		for idx := range futures {
+			// There is time difference between the workflow memory update and upload component inputs activity.
```
jvallesm, line 403: ❓ Does this mean that, if the component activity fails, we won't have access to the component input in the run logs? If so:
- I'd document that scenario here.
- This means that `ComponentActivity` depends on `UploadComponentInputsActivity` if we want the full log support, and that we shouldn't use futures for this activity. We can modify this post-release.

jvallesm (approved):
Indeed we should refactor this to reduce the code duplication. I added some non-blocking comments but you can take them as a reminder to myself for when I tackle this refactor.

pkg/utils/blobstorage.go
chuang8511: Oh, yes! I missed it! Thank you and the future you!

pkg/worker/workflow.go
chuang8511: > Does this mean that, if the component activity fails, we won't have access to the component input in the run logs?

No, it doesn't mean it.
It means `UploadComponentInputsActivity` should be executed after `ComponentActivity`.

Accurately, we should execute `UploadComponentInputsActivity` after the workflow memory is set.

However, because of unknown reason, probably worker execution sequence problem, it will be safer to execute `UploadComponentInputsActivity` after `UploadComponentOutputsActivity`.

I tested to set `UploadComponentInputsActivity` after `ComponentActivity`. But, somehow, the workflow memory sometimes did not set when I executed it after `ComponentActivity`.

jvallesm: Okay, then this should be investigated (otherwise the order of the activities is a bit puzzling) but I'm not against merging the current code.

chuang8511: > otherwise the order of the activities is a bit puzzling

Yes, I also want to know the reason! 
I might dig in Temporal document and do more experiments. 😃 

Thanks for pointing this out.

PR #939, opened 2024-12-16
feat(component,video): add task to embed audio to video

memoryonrepeat:
What this PR does:
- Implement a pipeline task to embed an audio input to a video input. The end result is the original video with audio stream being replaced by the provided audio.

CC @xiaofei-du @pinglin


go.mod
```diff
@@ -64,6 +64,7 @@ require (
 	github.com/slack-go/slack v0.12.5
 	github.com/tmc/langchaingo v0.1.10
 	github.com/u2takey/ffmpeg-go v0.5.0
+	github.com/warmans/ffmpeg-go v1.0.0
```
donch1989, line 67: Is there any specific reason for using a different ffmpeg-go package?

donch1989 (commented):
Hi @memoryonrepeat ,
Thanks for implementing this task. Could you add a few basic test cases first?
Thanks.

go.mod
memoryonrepeat: To implement this feature I had to call ffmpeg with more than one `map` parameter. The current ffmpeg-go library does not have support for that, other people also faced the same issue for a while, so someone ended up [implementing it](https://github.com/u2takey/ffmpeg-go/issues/1#issuecomment-2507904461). I also added the same [reference](https://github.com/instill-ai/pipeline-backend/pull/939/files#diff-6c50ab958e4a2e7d893ce563eea3c59d9fc3d21a86843b449baeb92cef73f324R96) in code.

memoryonrepeat:
@donch1989 I have added some test cases as well.

donch1989 (approved):

PR #952, opened 2025-01-14
refactor(pipeline): refactor the condition evaluation

donch1989:
Because

- The original implementation for pipeline condition evaluation is not robust. It converts the condition string into Go code and evaluates it using the AST. However, in our syntax, we typically use dot notation in recipes, while Go maps require bracket notation, especially when component or variable names are not valid Go identifiers.

This commit

- Refactors the condition evaluation.


jvallesm (approved):
⚡ ⚡  I couldn't have fixed this so quickly, I was still diving through the condition evaluation logic 😅

donch1989:
> ⚡ ⚡ I couldn't have fixed this so quickly, I was still diving through the condition evaluation logic 😅

The primary process involves converting our recipe syntax into Go code and utilizing the Go AST to evaluate the values. It may be somewhat complex to write initially, but I believe it won’t require much maintenance once it is stabilized.

PR #967, opened 2025-02-04
ci(github): use ARM runner to speed up the image build process

donch1989:
Because

- Originally, we used QEMU for cross-compiling images, but it was slow and unstable.

This commit

- Uses an ARM runner to speed up the image build process.


jvallesm (approved):
🙇

PR #981, opened 2025-03-04
feat(streaming): use Redis pub/sub to stream pipeline triggers

jvallesm:
Because

- On streamed pipeline triggers, the workers communicate with the API via in-process memory.
- This prevents us from separating the `main` and `worker` processes for more efficient scaling.

This commit

- Replaces the in-process memory communication by a Publication / Subscription pattern over Redis.

### DIFF
--- a/pkg/handler/triggerhandler.go
+++ b/pkg/handler/triggerhandler.go
@@ -845,10 +845,30 @@ func (sh *streamingHandler) Handle(ctx context.Context, triggerID string) {
 	}()
 
 	for {
-		event, err := sub.Receive(ctx)
-		if err != nil {
+		// Create a channel to receive the event asynchronously
+		eventCh := make(chan *memory.Event)
+		errCh := make(chan error)
+
+		// Start a goroutine to call sub.Receive
+		go func() {
+			event, err := sub.Receive(ctx)
+			if err != nil {
+				errCh <- err
+				return
+			}
+			eventCh <- event
+		}()
+
+		// Wait for either context cancellation or event received
+		var event *memory.Event
+		select {
+		case <-ctx.Done():
+			logger.Error("Context cancelled while waiting for event", zap.Error(ctx.Err()))
+			return
+		case err := <-errCh:
 			logger.Error("Couldn't receive message", zap.Error(err))
 			return
+		case event = <-eventCh:
 		}
 
 		if event.Name == string(memory.PipelineClosed) {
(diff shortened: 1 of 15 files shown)


jvallesm:
# QA ✅ 

https://github.com/user-attachments/assets/fb62eecf-6600-4d87-a52b-ab9a8f8b82c7

pkg/handler/triggerhandler.go
```diff
@@ -786,55 +813,62 @@ func request_PipelinePublicService_TriggerAsyncNamespacePipelineRelease_0_form(c
...
 	}
 }
 
 func (sh *streamingHandler) Handle(ctx context.Context, triggerID string) {
-	wfm, err := sh.ms.GetWorkflowMemory(ctx, triggerID)
-	if err != nil {
-		return
-	}
-	defer func() {
-		_ = sh.ms.PurgeWorkflowMemory(ctx, triggerID)
```
jvallesm, line 806: @donch1989 I removed this purge action here. Following the code, streaming endpoints call `TriggerAsync*` service usecases, so I added the same action [here](https://github.com/instill-ai/pipeline-backend/pull/981/files#diff-535874c672d3ba49bebb2add40574a260e061ba6df8286cd6c73eb3b8b51fe27R1851).

I noticed that there are many edge cases (mainly, non-happy paths) after the workflow memory has been created where it isn't cleared up. I didn't spend time fixing them, since we're moving away from the memory store.

pkg/memory/memory.go
```diff
@@ -83,8 +84,6 @@ type WorkflowMemory interface {
 
 	EnableStreaming()
 	IsStreaming() bool
-	SendEvent(_ context.Context, event *Event)
-	ListenEvent(_ context.Context) chan *Event
```
jvallesm, line 87: @donch1989 I'm thinking that maybe the removal of these methods and the `workflowMemory.channel` field allows us to move back the Temporal logic to the `worker` process. Do you see any potential blockers in this?

donch1989: I think it’s fine to remove them since they require the server and worker to be in the same process, which is not what we want.

donch1989: I think a missing piece is that the input data for a pipeline trigger is still [stored in workflow memory](https://github.com/instill-ai/pipeline-backend/blob/efea4ee7d7269ab3fcd9a2a5969b807da45bd7dd/pkg/service/pipeline.go#L925C23-L925C40), and the worker [retrieves it from the same memory](https://github.com/instill-ai/pipeline-backend/blob/efea4ee7d7269ab3fcd9a2a5969b807da45bd7dd/pkg/worker/workflow.go#L967-L968). This means the API server and the worker still need to be in the same process. We need to refactor this first before moving the Temporal worker back to the worker process.

jvallesm: Right, thanks for the pointer 👍

PR #986, opened 2025-03-18
feat(trigger): use MinIO to store workflow memory

jvallesm:
Because

- We're communicating the workflow memory information between the temporal clients and workers through an in-memory structure.
- This forces us to run the worker in the same process as the client.

This commit

- Stores and reads the workflow memory as a MinIO blob to decouple client and worker.

### DIFF
--- a/cmd/main/main.go
+++ b/cmd/main/main.go
@@ -299,7 +299,7 @@ func main() {
 	workerUID, _ := uuid.NewV4()
 
 	pubsub := pubsub.NewRedisPubSub(redisClient)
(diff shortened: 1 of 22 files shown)


jvallesm:
# QA ✅ 

- [x] Streaming and non-streaming pipeline (OpenAI)
- [x] Pipeline with iterator
- [x] Scheduled pipeline
- [ ] Pipeline with image inputs / outputs (url and binary)
- [ ] Overriding secrets and connections in the pipeline trigger

PR #987, opened 2025-03-19
feat(worker): extract worker back to its own process

jvallesm:
Because

- Now that client and worker are decoupled, we can use a separate process to run the Temporal workers, which will allow us to scale them independently.

This commit

- Moves back the Temporal worker logic to `cmd/worker`.
- Uses worker sessions to guarantee that activities share the workflow memory.
- Makes sure the worfklow memory (in-memory and in MinIO) is cleaned up.


jvallesm:
# QA ✅ 

Ran one server, 4 workers, and executed different pipelines

https://github.com/user-attachments/assets/c52bca2a-a6b0-40d8-baf5-6f9219e91d9c

PR #990, opened 2025-03-25
feat(worker): emit Temporal SDK metrics

jvallesm:
Because

- We want to observe Temporal metrics and scale workers according to them.

This commit

- Emits [Temporal Go SDK metrics](https://docs.temporal.io/develop/go/observability#metrics)


jvallesm:
## QA ✅ 
CleanShot 2025-03-25 at 18 46 09

PR #1007, opened 2025-05-12
ci(Makefile,workflows): update .env.secrets.component and integration test

pinglin:
Because

- `.env.secrets.component` has been adopted in other repos.

This commit

- makes `pipeline-backend` repo cohere with the rest.


pkg/handler/triggerhandler.go
```diff
@@ -532,7 +532,7 @@ func request_PipelinePublicService_TriggerAsyncNamespacePipeline_0(ctx context.C
 }
 
 // ref: the generated protogen-go files
-func request_PipelinePublicService_TriggerAsyncNamespacePipeline_0_form(ctx context.Context, marshaler runtime.Marshaler, client pb.PipelinePublicServiceClient, req *http.Request, pathParams map[string]string) (proto.Message, runtime.ServerMetadata, error) {
```
jvallesm, line 535: This is copied from the generated code in `protogen-go` but it doesn't seem to override the methods defined there (they're package methods that don't belong to any type) so this rename should be fine 👍 cc @donch1989

jvallesm (commented):
The changes enforced by the linter look good (and I thank you sincerely for bumping to v2 and wiping out all these if/else statements forever 😅 )

PR #1021, opened 2025-06-13
ci(workflow): update the dockerhub credentials in ga workflow

Sarthak-instill:
Because

- We have updated the login credentials from drop@instill.tech to drop@instill-ai.com in dockerhub.

This commit

- update the dockerhub credentials in ga workflow.


donch1989 (approved):

donch1989 (changes requested):
Those files should not be removed.

PR #1048, opened 2025-07-10
refactor(main): align backend codebase

pinglin:
Because

- we'd like to extract common components to share across all backends using the `x` package

This commit

- upgrades to Go 1.24.2
- adopts `x/log`
- replaces the `external` package with `x/client`
- removes tedious OTEL span and log codes (a more elegant middleware approach will be adopted)
- improves naming consistency


cmd/main/main.go
```diff
@@ -113,132 +90,22 @@ func main() {
...
-	}
+	publicGrpcS := grpc.NewServer(grpcServerOpts...)
+	reflection.Register(publicGrpcS)
 
-	artifactPrivateServiceClient, artifactPrivateServiceClientConn := external.InitArtifactPrivateServiceClient(ctx)
-	if artifactPrivateServiceClientConn != nil {
-		defer artifactPrivateServiceClientConn.Close()
-	}
+	// Initialize all clients
+	pipelinePublicServiceClient, mgmtPublicServiceClient, mgmtPrivateServiceClient,
```
jvallesm, line 103: I don't particularly love this way of initialising the clients. I've recently needed to modify the output of this script a couple of times (in a service with the similar pattern) and I had to kind of fish for the client I wanted to change. The same problem but for long argument lists is explained [here](https://google.github.io/styleguide/go/best-practices#function-argument-lists) quite clearly.

I saw nothing wrong in initialising each client in the main function but if we want to extract that logic to make it more readable and compact, I'd suggest returning a `clients` type (which can have a `close` method itself) to make the results more accessible.

pkg/handler/triggerhandler.go
```diff
@@ -812,8 +812,8 @@ func newStreamingHandler(writer http.ResponseWriter, sub pubsub.EventSubscriber)
 // TODO streamingHandler's methods should be merged into StreamingHandler as
 // unexported methods.
 func (sh *streamingHandler) handle(ctx context.Context, triggerID string) {
-	logger, _ := logger.GetZapLogger(ctx)
-	logger = logger.With(zap.String("triggerID", triggerID))
+	logger, _ := logx.GetZapLogger(ctx)
+	logger.Info("StreamingHandler", zap.String("triggerID", triggerID))
```
jvallesm, line 816: Any reason to log here? Also, this change removes the `triggerID` param from future logs in this method

pkg/usage/client.go
```diff
@@ -0,0 +1,41 @@
...
+	"google.golang.org/grpc/credentials/insecure"
+
+	"github.com/instill-ai/pipeline-backend/config"
+
+	usagepb "github.com/instill-ai/protogen-go/core/usage/v1beta"
+	logx "github.com/instill-ai/x/log"
+)
+
+// InitUsageServiceClient initializes a UsageServiceClient instance (no mTLS)
+func InitUsageServiceClient(ctx context.Context) (usagepb.UsageServiceClient, *grpc.ClientConn) {
```
jvallesm, line 19: 💡 If this is going to be the same in every service, we can extract it to `x` as `usage.NewClient(ctx, tlsEnabled, hostport)`

jvallesm (approved):

pkg/handler/triggerhandler.go
pinglin: This needs @donch1989 's input for the reason why logging here. I suspect it was some debug-purpose logging. It instantiated a new logger with the `triggerID` field by default but there is actually no further logs down the road.

PR #1076, opened 2025-09-01
fix(log): fix permission check for run logs

donch1989:
Because

- The permission check for private data is wrong.

This commit

- Adjusts the permission check for 
- Fixes wrong time unit for component logs

### DIFF
--- a/pkg/service/pipelinerun.go
+++ b/pkg/service/pipelinerun.go
@@ -100,7 +100,7 @@ func (s *service) ListPipelineRuns(ctx context.Context, req *pipelinepb.ListPipe
 
 	var referenceIDs []string
 	for _, pipelineRun := range pipelineRuns {
-		if canViewPrivateData(pipelineRun.RequesterUID, requesterUID) {
+		if canViewPrivateData(pipelineRun.RunnerUID, userUID) {
 			for _, input := range pipelineRun.Inputs {
 				referenceIDs = append(referenceIDs, input.Name)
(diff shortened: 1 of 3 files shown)


pkg/service/utils.go
```diff
@@ -159,7 +159,7 @@ func (s *service) convertComponentRunToPB(run datamodel.ComponentRun) (*pipeline
 	}
 
 	if run.CompletedTime.Valid {
-		totalDuration := int32(run.CompletedTime.Time.Sub(run.StartedTime))
+		totalDuration := int32(run.CompletedTime.Time.Sub(run.StartedTime).Milliseconds())
```
jvallesm, line 162: Good catch!

pkg/service/pipelinerun.go
```diff
@@ -100,16 +100,16 @@ func (s *service) ListPipelineRuns(ctx context.Context, req *pipelinepb.ListPipe
 
 	var referenceIDs []string
 	for _, pipelineRun := range pipelineRuns {
-		if canViewPrivateData(pipelineRun.RequesterUID, requesterUID) {
+		if canViewPrivateData(pipelineRun.RunnerUID, userUID) {
```
jvallesm, line 103: So this means that the run data will be visible when the authenticated user is the same that triggered the endpoint?

Do we want org runs to be visible for all the users within an organisation? I think that's why we had the requester UID check instead of the user UID

PR #1098, opened 2025-09-18
fix(usage): treat input rendering error as fatal

jvallesm:
Because

- Usage handler error produces a misleading error when the input data
  fails to be rendered.

This commit

- Fixes INS-8583
- Passes only the valid inputs to the usage checker.
- Returns a fatal error if no valid inputs could be rendered.


donch1989 (approved):

donch1989:
The unit test seems to be broken.

jvallesm:
> The unit test seems to be broken.

I broke a test by changing the error message, but it seems that there are other cases broken by a commit in the last couple of days 😖

PR #1100, opened 2025-09-18
fix(component,ai,instillmodel): resolve panics and test failures

pinglin:
Because

- The `instillmodel` component had critical panics due to nil interface conversions when system variables were missing or nil
- Tests were failing due to invalid base64 data, incorrect error message expectations, and hanging issues in image processing
- The model name parsing logic was vulnerable to index out of range errors when model names didn't follow the expected format

This commit

- Adds proper nil checks to `GetInstillUserUID` and `GetInstillRequesterUID` functions in helper.go to prevent interface conversion panics
- Implements model name format validation before splitting to prevent index out of range errors
- Fixes invalid base64 test data in text generation chat tests
- Updates test error message expectations to match actual implementation behavior
- Simplifies text-to-image output handling to resolve hanging issues and unmarshaling errors


jvallesm:
@pinglin please rebase 🙏

PR #1101, opened 2025-09-18
test(component,generic,http): replace external httpbin.org dependency with local test server

pinglin:
Because

- The unit tests were flaky, failing arbitrarily with 503 errors and connection timeouts
- Tests depended on external httpbin.org service, making them unreliable in CI/CD environments
- External network calls made tests slow (20+ seconds) and non-deterministic
- Type conversion panics occurred when httpbin.org returned unexpected response formats

This commit

- **Replaces httpbin.org with local httptest server**: Created `createLocalTestServer()` that mimics all httpbin.org endpoints (JSON, headers, text, binary, auth endpoints)
- **Eliminates external dependencies**: Tests now run completely offline with zero network calls to external services  
- **Adds test environment detection**: Implements `isTestEnvironment()` to cleanly bypass URL validation during testing without modifying production code
- **Improves test reliability**: Tests are now deterministic with consistent responses and no more flaky failures
- **Enhances performance**: Test execution time reduced from 20+ seconds to ~300ms
- **Maintains comprehensive coverage**: All HTTP methods, authentication types, and edge cases still fully tested
- **Fixes type conversion issues**: Added safe type checking to prevent panics when response format differs from expected
- **Enables parallel execution**: Each test gets its own server instance to avoid port conflicts

The refactored tests are now fast, reliable, and completely self-contained while maintaining full test coverage of the HTTP component functionality.

### DIFF
--- a/pkg/component/generic/http/v0/component_test.go
+++ b/pkg/component/generic/http/v0/component_test.go
@@ -2,7 +2,12 @@ package http
 
 import (
 	"context"
+	"encoding/json"
+	"io"
 	"net/http"
+	"net/http/httptest"
+	"os"
+	"strings"
 	"testing"
 
 	qt "github.com/frankban/quicktest"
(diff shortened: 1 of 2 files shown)


pkg/component/generic/http/v0/component_test.go
```diff
@@ -22,10 +27,211 @@ const (
...
+
+	return httptest.NewServer(mux)
+}
+
 func TestComponent(t *testing.T) {
 	c := qt.New(t)
 	c.Parallel()
 
+	// Set test environment to bypass URL validation
+	os.Setenv("GO_TESTING", "true")
```
jvallesm, line 232: Nit: instead of relying on env variables I'd made the whitelisted URLs or the URL validation function a constructor param for the component, it makes the behaviour of the component more explicit.

jvallesm (commented):
👍 much better with a server we can define and control

